PR #15, opened 2015-01-02
Update hash.js version in package.json

braydonf:


indutny:
What kind of issue does it cause?

braydonf:
If hash.js with version 0.3.2 is required separately, and npm-shrinkwrap is used to not duplicated the dependency:
```
npm WARN unmet dependency /var/www/bitpay/bitcore-tmp10/node_modules/bitcore/node_modules/elliptic requires hash.js@'^0.2.0' but will load
npm WARN unmet dependency /var/www/bitpay/bitcore-tmp10/node_modules/bitcore/node_modules/hash.js,
npm WARN unmet dependency which is version 0.3.2
```

indutny:
Released in 0.16.0, thank you.

PR #16, opened 2015-01-02
Bump `hash.js` to latest version

eordano:
This is generating npm issues when trying to use the latestest versions of `hash.js` and `elliptic` at the same time.


eordano:
LOL duplicated

Issue #17, opened 2015-01-04
point multiplication produces incorrect public key in rare situations


braydonf:
```javascript

var EC = require('elliptic').ec
var ec = new EC('secp256k1')

var point = ec.g.mul('f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: 41a2d4539b40a914e755a3a24367ce731195c354480dad4f2ce9e9c8f8003bce>

var point1 = ec.curve.pointFromX(true, '495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: 3dc9318f0a7ae59bafbccef98692931dc6a8c28208333e88de5c031766f70777>

var point2 = ec.curve.pointFromX(false, '495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: c236ce70f5851a6450433106796d6ce239573d7df7ccc17721a3fce79908f4b8>

var keypair = ec.keyPair('f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53')
// <Key priv: f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53 pub: null >

var pub = keypair.getPublic(true, 'hex')
// produces incorrect pub: '02495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba'

// correct values:
// pubkey: "03e275faa35bd1e88f5df6e8f9f6edb93bdf1d65f4915efc79fd7a726ec0c21700"
// address: "19FKem9YKbszEwsXRLw9WnkggMQrdLbAhE"
// wif: "L5MgSwNB2R76xBGorofRSTuQFd1bm3hQMFVf3u2CneFom8u1Yt7G"
```

Related: https://github.com/bitpay/bitcore/pull/894

indutny:
Got a shorter test:

```diff
[...]

PR #18, opened 2015-01-04
Tests: Added tests to reveal bug in public key calculation with precomputed values

braydonf:
Adds tests associated with issue #17


braydonf:
Removing line [425 and 426](https://github.com/indutny/elliptic/blob/master/lib/elliptic/curve/short.js#L425) fixes this, however benchmarks are slower.

indutny:
Thanks, landed other PR though.

PR #19, opened 2015-01-04
WIP: Problem with _endoWnafMulAdd

ryanxcharles:
There is a problem with point multiplication on short curves. If the point being multiplied does not have precomputed values, it can sometimes lead to incorrect results. This issue was discovered by @braydonf in https://github.com/indutny/elliptic/issues/17 . This WIP PR is an attempt to show explicitly what the problem is by adding a test that fails, and to fix it by removing the call to _endoWnafMulAdd. The failing test simply multiplies G by the same problematic value discovered by @braydonf, both when G does have precomputed values, and when G does not have precomputed values. This shouldn't make a difference to the result, but it does. I also demonstrate a solution, but the solution is inadequate because it makes point multiplication much slower. I'm not sure what the actual error is, other than that it is located somewhere in _endoWnafMulAdd.

### DIFF
--- a/test/curve-test.js
+++ b/test/curve-test.js
@@ -74,6 +74,18 @@ describe('Curve', function() {
       k.toString(16));
   });
 
+  it('should compute this problematic secp256k1 multiplication', function() {
+    var curve = elliptic.curves.secp256k1.curve;
+    var g1 = curve.g; // precomputed g
+    assert(g1.precomputed);
+    var g2 = curve.point(g1.getX(), g1.getY()); // not precomputed g
+    assert(!g2.precomputed);
+    var a = new bn('6d1229a6b24c2e775c062870ad26bc261051e0198c67203167273c7c62538846', 16);
+    var p1 = g1.mul(a);
+    var p2 = g2.mul(a);
+    assert(p1.eq(p2));
+  });
+
   it('should not fail on secp256k1 regression', function() {
     var curve = elliptic.curves.secp256k1.curve;
     var k1 = new bn('32efeba414cd0c830aed727749e816a01c471831536fd2fce28c56b54f5a3bb1', 16);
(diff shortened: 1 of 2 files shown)


braydonf:
Interesting, I tried removing the precomputed statement (just above on line 125 ` lib/elliptic/curve/short.js `) and was able to get the correct values. 

I just added a check for the correct x and y values, and it also works:
```javascript
assert(p1.eq(p2));
assert.equal(p1.getX().toString(16), 'd6106302d2698d6a41e9c9a114269e7be7c6a0081317de444bb2980bf9265a01');
assert.equal(p1.getY().toString(16), 'e05fb262e64b108991a29979809fcef9d3e70cafceb3248c922c17d83d66bc9d');
````

File: test/curve-test.js [hunk 1]
```diff
@@ -74,6 +74,18 @@ describe('Curve', function() {
...
+  it('should compute this problematic secp256k1 multiplication', function() {
+    var curve = elliptic.curves.secp256k1.curve;
+    var g1 = curve.g; // precomputed g
+    assert(g1.precomputed);
+    var g2 = curve.point(g1.getX(), g1.getY()); // not precomputed g
+    assert(!g2.precomputed);
+    var a = new bn('6d1229a6b24c2e775c062870ad26bc261051e0198c67203167273c7c62538846', 16);
+    var p1 = g1.mul(a);
+    var p2 = g2.mul(a);
+    assert(p1.eq(p2));
```
braydonf: adding a check for the values here also will also pass:
```javascript
assert.equal(p1.getX().toString(16), 'd6106302d2698d6a41e9c9a114269e7be7c6a0081317de444bb2980bf9265a01');
assert.equal(p1.getY().toString(16), 'e05fb262e64b108991a29979809fcef9d3e70cafceb3248c922c17d83d66bc9d');
```

indutny:
Looking.

indutny:
The problem is in `_wnafMulAdd`, looking more deeply into it.

indutny:
Well at least, if I'll replace `_wnafMulAdd` in `_endoWnafMulAdd` with two `_wnafMul` and one `add` - it'll work.

indutny:
Here is a shorter test: https://github.com/indutny/elliptic/issues/17#issuecomment-68628830

indutny:
Partially landed, thank you.

PR #20, opened 2015-01-12
don't bring in the whole assert library

calvinmetcalf:
I grabbed the assert function you used in bn.js, considering you only use assert() assert by itself is pretty big PLUS it brings in utils

### DIFF
--- a/lib/elliptic/curve/base.js
+++ b/lib/elliptic/curve/base.js
@@ -1,7 +1,6 @@
 var bn = require('bn.js');
 var elliptic = require('../../elliptic');
 
-
 var getNAF = elliptic.utils.getNAF;
 var getJSF = elliptic.utils.getJSF;
 var assert = elliptic.utils.assert;
--- a/lib/elliptic/curve/base.js
+++ b/lib/elliptic/curve/base.js
@@ -1,9 +1,10 @@
-var assert = require('assert');
 var bn = require('bn.js');
 var elliptic = require('../../elliptic');
 
+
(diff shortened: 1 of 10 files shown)


lib/elliptic/utils.js
```diff
@@ -1,8 +1,10 @@
-var assert = require('assert');
 var bn = require('bn.js');
 
 var utils = exports;
-
```
indutny: Please do not remove newline here.

lib/elliptic/utils.js
```diff
@@ -1,8 +1,10 @@
...
+utils.assert = function assert(val, msg) {
+  if (!val)
+    throw new Error(msg || 'Assertion failed');
+};
```
indutny: Please add a newline after this line.

lib/elliptic/hmac-drbg.js
```diff
@@ -1,8 +1,8 @@
-var assert = require('assert');
```
indutny: I guess there is no need in this newline, right?

lib/elliptic/curve/base.js
```diff
@@ -1,9 +1,10 @@
-var assert = require('assert');
 var bn = require('bn.js');
 var elliptic = require('../../elliptic');
 
+
```
indutny: Superfluous newline.

indutny:
Thanks! Generally LGTM, except some style nits.

calvinmetcalf:
fixed!

indutny:
Landed!

PR #22, opened 2015-01-16
key: explicit methods for priv/pub/object

### LINKED ISSUE #21: query: curve25519-donna
i have been looking around for a pure-javascript implementation of the "donna" -- https://code.google.com/p/curve25519-donna/ & http://cr.yp.to/ecdh.html

i have a sneaking suspicion that this is implemented in elliptic, although with a slightly different API/naming convention. am i right or just optimistic?

sorry for the lack of clue. thanks!

indutny:
fix #21

@ryanxcharles how does this look to you guys? Should remove lots of uncertainty from the code.

### DIFF
--- a/lib/elliptic/curve/edwards.js
+++ b/lib/elliptic/curve/edwards.js
@@ -13,7 +13,7 @@ function EdwardsCurve(conf) {
   this.mOneA = this.twisted && conf.a == -1;
   this.extended = this.mOneA;
 
-  Base.call(this, 'mont', conf);
+  Base.call(this, 'edwards', conf);
 
   this.a = new bn(conf.a, 16).mod(this.red.m).toRed(this.red);
   this.c = new bn(conf.c, 16).toRed(this.red);
--- a/lib/elliptic/curve/mont.js
+++ b/lib/elliptic/curve/mont.js
@@ -25,7 +25,7 @@ MontCurve.prototype.point = function point(x, z) {
 
 MontCurve.prototype.pointFromJSON = function pointFromJSON(obj) {
   return Point.fromJSON(this, obj);
-}
+};
 
 MontCurve.prototype.validate = function validate(point) {
   var x = point.normalize().x;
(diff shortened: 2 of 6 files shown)


mrose17:
+1

dcousens:
+1 I'm very happy with these changes.

jprichardson:
Yeah, this definitely makes it much nicer to work with!

indutny:
cc @braydonf too

braydonf:
LGTM

I've tested these changes with [Bitcore](https://github.com/bitpay/bitcore) and it can upgraded without any changes. With [BitAuth](https://github.com/bitpay/bitauth) two lines would need to be changed:

```patch
@@ -33,11 +33,11 @@ BitAuth.generateSin = function() {
  * @param {String} A private key in hex
  * @returns {String} A compressed public key in hex
  */
 BitAuth.getPublicKeyFromPrivateKey = function(privkey) {
 
-  var keys = ecdsa.keyPair(privkey, 'hex');
+  var keys = ecdsa.keyFromPrivate(privkey, 'hex');
 
   // compressed public key
   var pubKey = keys.getPublic();
   var xbuf = new Buffer(pubKey.x.toString('hex', 64), 'hex');
   var ybuf = new Buffer(pubKey.y.toString('hex', 64), 'hex');
@@ -112,11 +112,11 @@ BitAuth.sign = function(data, privkey) {
  * @returns {Function|Boolean} - If the signature is valid
  */
 BitAuth.verifySignature = function(data, pubkey, hexsignature, callback) {
   var hash = (new hashjs.sha256()).update(data).digest('hex');
   var signature = new Buffer(hexsignature, 'hex');
-  var valid = ecdsa.verify(hash, signature, pubkey);
+  var valid = ecdsa.verify(hash, signature, pubkey, 'hex');
   if (callback)
     return callback(null, valid);
   return valid;
 };
 ```

It looks like the format of the pubkey in verify was assumed to be 'hex' before, and now it needs to be explicitly set.

indutny:
Landed and pushed in 2.0.0, because of the API change.

mrose17:
+1 - many thanks!

PR #26, opened 2015-01-21
Swap out brorand and hash.js

calvinmetcalf:
this swaps out brorand and hash.js for randombytes, create-hash, and create-hmac, all 3 of the libraries are the modularized crypto browserify shims which just use the crypto module in node and only use the pure js shim in the browser.

### DIFF
--- a/lib/elliptic/hmac-drbg.js
+++ b/lib/elliptic/hmac-drbg.js
@@ -1,4 +1,4 @@
-var hash = require('hash.js');
+var hash = require('./hash');
 var elliptic = require('../elliptic');
 var utils = elliptic.utils;
 var assert = utils.assert;
@@ -27,10 +27,10 @@ function HmacDRBG(options) {
 module.exports = HmacDRBG;
 
 HmacDRBG.prototype._init = function init(entropy, nonce, pers) {
-  var seed = entropy.concat(nonce).concat(pers);
+  var seed = new Buffer(entropy.concat(nonce).concat(pers));
 
-  this.K = new Array(this.outLen / 8);
-  this.V = new Array(this.outLen / 8);
+  this.K = new Buffer(this.outLen / 8);
+  this.V = new Buffer(this.outLen / 8);
   for (var i = 0; i < this.V.length; i++) {
     this.K[i] = 0x00;
     this.V[i] = 0x01;
@@ -48,17 +48,16 @@ HmacDRBG.prototype._hmac = function hmac() {
 HmacDRBG.prototype._update = function update(seed) {
   var kmac = this._hmac()
                  .update(this.V)
-                 .update([ 0x00 ]);
+                 .update(new Buffer([0x00]));
   if (seed)
     kmac = kmac.update(seed);
   this.K = kmac.digest();
   this.V = this._hmac().update(this.V).digest();
   if (!seed)
     return;
-
   this.K = this._hmac()
                .update(this.V)
-               .update([ 0x01 ])
+               .update(new Buffer([0x01]))
                .update(seed)
                .digest();
   this.V = this._hmac().update(this.V).digest();
@@ -72,8 +71,8 @@ HmacDRBG.prototype.reseed = function reseed(entropy, entropyEnc, add, addEnc) {
     entropyEnc = null;
   }
 
-  entropy = utils.toBuffer(entropy, entropyEnc);
-  add = utils.toBuffer(add, addEnc);
+  entropy = utils.toArray(entropy, entropyEnc);
+  add = utils.toArray(add, addEnc);
 
   assert(entropy.length >= (this.minEntropy / 8),
          'Not enough entropy. Minimum is: ' + this.minEntropy + ' bits');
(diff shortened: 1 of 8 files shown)


lib/elliptic/hmac-drbg.js
```diff
@@ -27,10 +27,10 @@ function HmacDRBG(options) {
 module.exports = HmacDRBG;
 
 HmacDRBG.prototype._init = function init(entropy, nonce, pers) {
-  var seed = entropy.concat(nonce).concat(pers);
+  var seed = new Buffer(entropy.concat(nonce).concat(pers));
 
-  this.K = new Array(this.outLen / 8);
-  this.V = new Array(this.outLen / 8);
+  this.K = new Buffer(this.outLen / 8);
```
indutny: Dependency on `Buffer`? Why?

indutny:
The whole elliptic module is used for browsers, right? Why bother with node-only dependencies?

lib/elliptic/hmac-drbg.js
calvinmetcalf: node hash functions can't handle arrays of numbers

calvinmetcalf:
so when used in node it uses crypto and when used in the browser it uses pieces from crypto browserify, that lets me make a createECDH shim that works in node 0.10, node 0.11/io.js and the browser using the  native hashes if available.

calvinmetcalf:
so I approached this from the perspective of wanting to avoid duplicate hash libraries in crypto-browserify, i can probably do this in a way that avoids adding buffers as a dependency by using sha.js directly

indutny:
Yeah, I think this would be more preferable. I really do enjoy the minimal footprint that this library has right now.

calvinmetcalf:
rats sha.js is also buffer based

dcousens:
+1 really happy with the direction of this.

When I use elliptic.js right now I end up with a much bigger foot print simply because I have to pull in both sets of hash libraries.

calvinmetcalf:
Yeah I've been giving this more thought and it's going to be hard to make
crypto-browserify smaller without making this library bigger

dcousens:
`Buffer.js` is ~1000 SLOC,   take away all the buffer-like utility functions in elliptic and replace them to all use buffer.js,  and you're left with ~5-600 lines more than it has now.

Yes, it will make this library bigger, but it also opens up the library to a consistent interface AS well as dramatically reducing any library that uses both.

Given how popular `Buffer` is in most other libraries, let alone if you touch anything with streams,  I'd wager for any non-toy example, you will see a dramatic decrease in net size.

Kagami:
+1 for replacing brorand because webpack is too smart and require entire `crypto-browserify` package because of this line: https://github.com/indutny/brorand/blob/master/index.js#L43
Also +1 for replacing `hash.js` with `sha.js` because `sha.js` has better performance for SHA-512 and `sha.js` is much more common so it would result in smaller browserify build sizes.

indutny:
@calvinmetcalf I'm trying hard to convince myself to merge this :) If you don't mind please give me some help here. How does it affect performance? I know that this repo does not have any benchmarks yet, but if you could figure something out (both io.js and browser) - it would be super cool.

P.S. You may try something like http://jsperf.com/jsbn-vs-sjcl-ecc as a starting point.

Thank you!

calvinmetcalf:
http://jsperf.com/jsbn-vs-sjcl-ecc/21 hash.js is significantly faster then sha.js even after optimizing the hash and hmac functions to avoid streams being invoked
screen shot 2015-01-26 at 2 49 59 pm

if you compare the 2 cals to getKeyPare the hash update calls in hash.js are significantly less costly, we should probably look into incorperating these changes into sha.js or just using it instead of sha.js

indutny:
@calvinmetcalf thank you! I guess this is not ready to land then?

dcousens:
I'll look into the performance issues.

calvinmetcalf:
@indutny yeah maybe scale this back to just replace brorand with randombytes

indutny:
@calvinmetcalf could I be added as an owner of randombytes? I'd like to have some sort of control over my crypto stuff :)

indutny:
Or... I have even better solution! Why not make two optional dependencies: `brorand` and/or `randombytes`. Then assert that one of them is present. What do you think?

Kagami:
>Why not make two optional dependencies: brorand and/or randombytes. Then assert that one of them is present.

Then browserify will include both of the into the build (if they required inside the js file), won't it?

calvinmetcalf:
@dcousens we probably want to hit sha.js directly from create-hmac, that would save a lot of indirection

@indutny invited you to join crypto-browerify

indutny:
Oh right! I totally forgot about the invite, thanks! What about `npm`?

calvinmetcalf:
Add yourself to the owners repo

jprichardson:
@indutny 

Add yourself to this file: https://github.com/crypto-browserify/ownership/blob/master/crypto-users.json

indutny:
Thanks! Done.

calvinmetcalf:
@indutny added

indutny:
Yay, thanks!

dcousens:
@calvinmetcalf would you be able to isolate those performance cases?

calvinmetcalf:
@dcousens it comes down to the speed of making a hash so it's really just

```js
createHash('sha256').update('something').digest();
```

vs

```js
hash.sha256().update('something').digest();
```

indutny:
Looks like it didn't lift off. Sorry!

dcousens:
@indutny maybe another day :+1:

fanatid:
I hope using Buffer, [sha.js](https://www.npmjs.com/package/sha.js), [randombytes](https://www.npmjs.com/package/randombytes) in plans (here and bn.js)?
because crypto-browserify uses elliptic, bn.js and they brings extra dependencies..

indutny:
@fanatid if you insist :wink:

PR #27, opened 2015-01-23
Added support for ECIES with curve ed25519

timtutt:
Adding support for ECIES. Making Use of CryptoJS AES library for random
IV, Salt generation and AES 256. Updated Readme with example. Added
ecies-test.js


lib/cryptojs/aes.js
```diff
@@ -0,0 +1,37 @@
+/*
```
indutny: I wonder if it could be installed using npm dependencies?

indutny:
Oh, actually! I just realized that ECIES support does not require any changes to the library's core, am I right?

If so - I think it will be much better to publish it as a separate module! What are your thoughts on this?

Kagami:
Btw, example of ECIES implementation using `elliptic` and `WebCryptoAPI`: https://github.com/bitchan/eccrypto/blob/531148466283acfa1a464832149d833381246c2c/browser.js#L113-L141

indutny:
Ok, @timtutt thank you very much for contribution, but I think it would be much more useful for everyone to publish it as a separate module with dependency to `elliptic`. Please do not hesitate to post a link here or open a PR for README.md with a link to your project. (@Kagami I'd be pleased to accept same PR from you ;) )

timtutt:
@indutny - Sure makes sense to me. Will shoot a PR with the changes pulled
out into a separate module later today.

PR #28, opened 2015-01-26
Add link to eccrypto per #27

Kagami:


indutny:
Landed in 99c5d0d, thank you!

PR #29, opened 2015-02-08
Add dev dependencies needed for benchmarking

weilu:


dcousens:
ACK

indutny:
Hm... Ain't they present in `benchmark/package.json`?

weilu:
I see that now. I'm used to `npm install` and expect everything needed for dev installed. I guess this project has a different convention.

PR #31, opened 2015-02-14
Fixed PK encoding unused during import

alexeykudinkin:
Typo?

### DIFF
--- a/lib/elliptic/ec/key.js
+++ b/lib/elliptic/ec/key.js
@@ -93,7 +93,7 @@ KeyPair.prototype.getPrivate = function getPrivate(enc) {
 };
 
 KeyPair.prototype._importPrivate = function _importPrivate(key, enc) {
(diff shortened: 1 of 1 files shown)


indutny:
Maybe... But why are the tests failing?

lib/elliptic/ec/key.js
```diff
@@ -93,7 +93,7 @@ KeyPair.prototype.getPrivate = function getPrivate(enc) {
 };
 
 KeyPair.prototype._importPrivate = function _importPrivate(key, enc) {
-  this.priv = new bn(key, 16);
+  this.priv = new bn(key, enc);
```
indutny: May be `enc || 16`?

alexeykudinkin:
Yeah, you're right. 
Sorry for not firing tests before PR.

indutny:
oh, forgot to say. Landed, thank you!

PR #37, opened 2015-03-31
init stab at recover signatures

wanderer:
WIP not ready

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -179,7 +179,6 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
     throw new Error('Unable to find sencond key candinate');
 
   // 1.1. Let x = r + jn.
-  // var x = isSecondKey ? n.add(r) : r;
   r = this.curve.pointFromX(isYOdd, r);
   var eNeg = e.neg().mod(n);
 
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -166,7 +166,6 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 };
 
 EC.prototype.recoverPubKey = function(msg, signature, j) {
-
   assert(3 & j === j, 'The recovery param is more than two bits');
   var n = this.n;
   var e = new bn(msg);
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -173,6 +173,7 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 EC.prototype.recoverPubKey = function(msg, signature, j) {
   var n = this.n;
   var g = this.g;
+  var e = new bn(msg);
 
   var r = signature.r;
   var s = signature.s;
@@ -186,25 +187,19 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
 
   //1.1. Let x = r + jn.
   var x = isSecondKey ? n.add(r) : r;
-  // console.log('x: ' + x.toString());
 
-  var r = this.curve.pointFromX(isYOdd, x);
-
-  //mul doesn't work
+  r = this.curve.pointFromX(isYOdd, x);
   var rn = r.mul(n);
-  console.log(rn.isInfinity(rn));
-  
-  
-  assert(rn.isInfinity(rn), 'nR is not a valid curve point')
-  
-  // var eNeg = a.neg().mod(n);
+
+  assert(rn.isInfinity(), 'nR is not a valid curve point');
+
+  var eNeg = e.neg().mod(n);
 
   // 1.6.1 Compute Q = r^-1 (sR -  eG)
   //               Q = r^-1 (sR + -eG)
-  // var rInv = r.modInverse(n)
+  var rInv = signature.r.invm(n);
 
-  // var Q = R.multiplyTwo(s, G, eNeg).multiply(rInv)
-  // curve.validate(Q)
-
-  // return Q
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+  assert(this.curve.validate(Q), 'Invalid Point');
+  return Q;
 };
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -174,18 +174,13 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
 
   // A set LSB signifies that the y-coordinate is odd
   var isYOdd = j & 1;
-
-  // The more significant bit specifies whether we should use the
-  // first or second candidate key.
   var isSecondKey = j >> 1;
+  if (r.cmp(this.curve.p.mod(this.curve.n)) >= 0 && isSecondKey)
+    throw new Error('Unable to find sencond key candinate');
 
   // 1.1. Let x = r + jn.
-  var x = isSecondKey ? n.add(r) : r;
-  r = this.curve.pointFromX(isYOdd, x);
-
-  var rn = r.mul(n);
-  assert(rn.isInfinity(), 'nR is not a valid curve point');
-
+  // var x = isSecondKey ? n.add(r) : r;
+  r = this.curve.pointFromX(isYOdd, r);
   var eNeg = e.neg().mod(n);
 
   // 1.6.1 Compute Q = r^-1 (sR -  eG)
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -173,9 +173,7 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 EC.prototype.recoverPubKey = function(msg, signature, j) {
   j = new bn(j);
   var n = this.n;
-  var g = this.g;
   var e = new bn(msg);
-
   var r = signature.r;
   var s = signature.s;
 
@@ -186,7 +184,7 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
   // first or second candidate key.
   var isSecondKey = j >> 1;
 
-  //1.1. Let x = r + jn.
+  // 1.1. Let x = r + jn.
   var x = isSecondKey ? n.add(r) : r;
 
   r = this.curve.pointFromX(isYOdd, x);
@@ -213,5 +211,5 @@ EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
       return i;
     }
   }
-  throw new Error('Unable to find valid recovery factor')
+  throw new Error('Unable to find valid recovery factor');
 };
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -193,15 +193,16 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
   var rInv = signature.r.invm(n);
 
   var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
-  assert(this.curve.validate(Q), 'Invalid Point');
   return Q;
 };
 
-EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
+EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q, enc) {
+  signature = new Signature(signature, enc);
   for (var i = 0; i < 4; i++) {
     var Qprime = this.recoverPubKey(e, signature, i);
 
-    if (Qprime.eq(Q))  return i;
+    if (Qprime.eq(Q))
+      return i;
   }
   throw new Error('Unable to find valid recovery factor');
 };
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -21,9 +21,7 @@ function EC(options) {
 
   // Shortcut for `elliptic.ec(elliptic.curves.curveName)`
   if (options instanceof elliptic.curves.PresetCurve)
-    options = {
-      curve: options
-    };
+    options = { curve: options };
 
   this.curve = options.curve.curve;
   this.n = this.curve.n;
@@ -138,10 +136,7 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (options.canonical && s.cmp(this.nh) > 0)
       s = this.n.sub(s);
 
-    return new Signature({
-      r: r,
-      s: s
-    });
+    return new Signature({ r: r, s: s });
   } while (true);
 };
 
@@ -171,7 +166,8 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 };
 
 EC.prototype.recoverPubKey = function(msg, signature, j) {
-  j = new bn(j);
+
+  assert(3 & j === j, 'The recovery param is more than two bits');
   var n = this.n;
   var e = new bn(msg);
   var r = signature.r;
@@ -186,10 +182,9 @@ EC.prototype.recoverPubKey = function(msg, signature, j) {
 
   // 1.1. Let x = r + jn.
   var x = isSecondKey ? n.add(r) : r;
-
   r = this.curve.pointFromX(isYOdd, x);
-  var rn = r.mul(n);
 
+  var rn = r.mul(n);
   assert(rn.isInfinity(), 'nR is not a valid curve point');
 
   var eNeg = e.neg().mod(n);
@@ -207,9 +202,7 @@ EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
   for (var i = 0; i < 4; i++) {
     var Qprime = this.recoverPubKey(e, signature, i);
 
-    if (Qprime.eq(Q)) {
-      return i;
-    }
+    if (Qprime.eq(Q))  return i;
   }
   throw new Error('Unable to find valid recovery factor');
 };
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -190,7 +190,7 @@ EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
   return r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
 };
 
-EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q, enc) {
+EC.prototype.getKeyRecoveryParam = function(e, signature, Q, enc) {
   signature = new Signature(signature, enc);
   for (var i = 0; i < 4; i++) {
     var Qprime = this.recoverPubKey(e, signature, i);
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -166,8 +166,9 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 };
 
 EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
-  assert(3 & j === j, 'The recovery param is more than two bits');
+  assert((3 & j) === j, 'The recovery param is more than two bits');
   signature = new Signature(signature, enc);
+
   var n = this.n;
   var e = new bn(msg);
   var r = signature.r;
@@ -186,8 +187,7 @@ EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
   // 1.6.1 Compute Q = r^-1 (sR -  eG)
   //               Q = r^-1 (sR + -eG)
   var rInv = signature.r.invm(n);
-  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
-  return Q;
+  return r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
 };
 
 EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q, enc) {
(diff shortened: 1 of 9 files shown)


lib/elliptic/ec/index.js
```diff
@@ -164,3 +169,42 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  var isSecondKey = j >> 1;
+
+  //1.1. Let x = r + jn.
+  var x = isSecondKey ? n.add(r) : r;
+  // console.log('x: ' + x.toString());
+
+  var r = this.curve.pointFromX(isYOdd, x);
+
+  //mul doesn't work
+  var rn = r.mul(n);
```
wanderer: @indutny there appears to be a problem with Point.mul here. some points on the curver appear to not be in red when they should.  Using this tests case https://gist.github.com/wanderer/d1726fea77e5f2a8d46d

wanderer: /home/null/code/elliptic/node_modules/bn.js/lib/bn.js:9
    throw new Error(msg || 'Assertion failed');
    ^
Error: red works only with positives
    at assert (/home/null/code/elliptic/node_modules/bn.js/lib/bn.js:9:11)
    at Red._verify2 (/home/null/code/elliptic/node_modules/bn.js/lib/bn.js:2021:3)
    at Red.isub (/home/null/code/elliptic/node_modules/bn.js/lib/bn.js:2066:8)
    at BN.redISub (/home/null/code/elliptic/node_modules/bn.js/lib/bn.js:1755:19)
    at Point.add (/home/null/code/elliptic/lib/elliptic/curve/short.js:388:39)
    at ShortCurve._wnafMulAdd (/home/null/code/elliptic/lib/elliptic/curve/base.js:153:27)
    at ShortCurve._endoWnafMulAdd (/home/null/code/elliptic/lib/elliptic/curve/short.js:240:18)
    at Point.mul (/home/null/code/elliptic/lib/elliptic/curve/short.js:427:23)
    at EC.recoverPubKey (/home/null/code/elliptic/lib/elliptic/ec/index.js:194:14)
    at Object.<anonymous> (/home/null/code/elliptic/test.js:25:4)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)

indutny: @wanderer why are you using BigInteger in your test code?

wanderer: because im testing against the `ecdsa` module which use BigInterger. BigInterger is not being used with any with any elliptics's code

indutny: Oh gosh, sorry then. I'll look into it this week.

indutny: Sorry, @wanderer , took too long for me. Fixed in bn.js@2.0.5, please give it another spin if you still have time ;)

wanderer:
@indutny the recoveryPubKey function now works. I also added test. Please review when you get a chance.

dcousens:
Seeing as your implementation is taken directly from the implementation in [bitcoinjs-lib](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecdsa.js#L147-L219),  you could definitely improve the amount of testing you are merging by also just copying [the tests verbatim](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/ecdsa.js#L87-L137).

sublimator:
Great, we need this functionality too

wanderer:
@dcousens Thanks! ill try those tests. I actually ripped the implementation [from here](https://github.com/cryptocoinjs/ecdsa/blob/master/lib/ecdsa.js#L115-L151) but it looks close to bitcoinjs as well.

lib/elliptic/ec/index.js
```diff
@@ -21,7 +21,9 @@ function EC(options) {
 
   // Shortcut for `elliptic.ec(elliptic.curves.curveName)`
   if (options instanceof elliptic.curves.PresetCurve)
-    options = { curve: options };
+    options = {
```
indutny: Why?

lib/elliptic/ec/index.js
```diff
@@ -136,7 +138,10 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (options.canonical && s.cmp(this.nh) > 0)
       s = this.n.sub(s);
 
-    return new Signature({ r: r, s: s });
+    return new Signature({
```
indutny: Ok... but why?

lib/elliptic/ec/index.js
```diff
@@ -164,3 +169,47 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+
+EC.prototype.recoverPubKey = function(msg, signature, j) {
+  j = new bn(j);
+  var n = this.n;
+  var e = new bn(msg);
+  var r = signature.r;
+  var s = signature.s;
+
+  // A set LSB signifies that the y-coordinate is odd
+  var isYOdd = j & 1;
```
indutny: `j.isOdd()`?

lib/elliptic/ec/index.js
```diff
@@ -164,3 +169,47 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+
+  // The more significant bit specifies whether we should use the
+  // first or second candidate key.
+  var isSecondKey = j >> 1;
```
indutny: I think it is pretty odd to do the regular arithmetic ops on BN instances. What do you think about doing `j.shrn(1)`?

indutny: Or better `j.testn(1)`

lib/elliptic/ec/index.js
```diff
@@ -164,3 +169,47 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+  assert(this.curve.validate(Q), 'Invalid Point');
+  return Q;
+};
+
+EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
+  for (var i = 0; i < 4; i++) {
+    var Qprime = this.recoverPubKey(e, signature, i);
+
+    if (Qprime.eq(Q)) {
```
indutny: No need for braces.

lib/elliptic/ec/index.js
indutny: Btw, are you sure that you are checking MSB with this `>> 1` thing? I guess you need something like `j.testn(j.bitLength() - 1)` to actually do it.

indutny: Or if you are checking the sign - you may want use `this.n.bitLength() - 1`

lib/elliptic/ec/index.js
wanderer: text editor formatted that, ill revert there.

lib/elliptic/ec/index.js
wanderer: this shouldn't be a BN, should just be a Number. Will change line 174

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,45 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  return Q;
+};
+
+EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
+  for (var i = 0; i < 4; i++) {
+    var Qprime = this.recoverPubKey(e, signature, i);
+
+    if (Qprime.eq(Q))  return i;
+  }
+  throw new Error('Unable to find valid recovery factor');
```
wanderer: This is slow. But I don't know the correct way to do this. Bitcoin generates the recovery ID on when signing. See [here](https://github.com/bitcoin/secp256k1/blob/master/src/ecdsa_impl.h#L223-L260) I'm not sure what the `overflow` is doing yet.  So I haven't attempted at porting it.

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,45 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 
   return p.getX().mod(this.n).cmp(r) === 0;
 };
+
+EC.prototype.recoverPubKey = function(msg, signature, j) {
+
```
indutny: Superfluous newline ;)

lib/elliptic/ec/key.js
```diff
@@ -16,6 +16,7 @@ function KeyPair(ec, options) {
   if (options.pub)
     this._importPublic(options.pub, options.pubEnc);
 }
+
```
indutny: No need in changing this ;)

lib/elliptic/ec/index.js
wanderer: removed now

lib/elliptic/ec/key.js
wanderer: fixed!

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,44 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  // The more significant bit specifies whether we should use the
+  // first or second candidate key.
+  var isSecondKey = j >> 1;
+
+  // 1.1. Let x = r + jn.
+  var x = isSecondKey ? n.add(r) : r;
+  r = this.curve.pointFromX(isYOdd, x);
+
+  var rn = r.mul(n);
+  assert(rn.isInfinity(), 'nR is not a valid curve point');
```
indutny: I'd rather not do this assert. AFAIK, It is possible easy to get the point that is not on curve when nudging `r` with `curve.n`.

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,44 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+
+  // A set LSB signifies that the y-coordinate is odd
+  var isYOdd = j & 1;
+
+  // The more significant bit specifies whether we should use the
+  // first or second candidate key.
+  var isSecondKey = j >> 1;
+
+  // 1.1. Let x = r + jn.
+  var x = isSecondKey ? n.add(r) : r;
```
indutny: A suggestion. There can't be a second key if `r` is bigger or equal to `curve.p.mod(curve.n)`.

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,44 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  assert(rn.isInfinity(), 'nR is not a valid curve point');
+
+  var eNeg = e.neg().mod(n);
+
+  // 1.6.1 Compute Q = r^-1 (sR -  eG)
+  //               Q = r^-1 (sR + -eG)
+  var rInv = signature.r.invm(n);
+
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+  assert(this.curve.validate(Q), 'Invalid Point');
```
indutny: This can't really be the case, right?

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,44 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 
   return p.getX().mod(this.n).cmp(r) === 0;
 };
+
+EC.prototype.recoverPubKey = function(msg, signature, j) {
```
indutny: Should do the `new Signature(...)` trick, and should have an encoding option for it.

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,44 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+  assert(this.curve.validate(Q), 'Invalid Point');
+  return Q;
+};
+
+EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q) {
+  for (var i = 0; i < 4; i++) {
+    var Qprime = this.recoverPubKey(e, signature, i);
+
+    if (Qprime.eq(Q))  return i;
```
indutny: Please put `return i;` on the next line.

lib/elliptic/ec/index.js
wanderer: Not sure what you mean here.  Its following the spec though. [on page 54](http://www.secg.org/sec1-v2.pdf)
`1.4. If nR 6= O, then do another iteration of Step 1.`
I think this applies here.

lib/elliptic/ec/index.js
wanderer: sure, removed!

lib/elliptic/ec/index.js
indutny: Yeah, but this code is throwing, so you won't be doing a next iteration anymore ;)

lib/elliptic/ec/index.js
wanderer: im not sure what to do with that.

indutny: You may skip this check altogether and throw if `r` is bigger than that value and `isSecondKey` is `true`.

lib/elliptic/ec/index.js
wanderer: done!

lib/elliptic/ec/index.js
wanderer: added!

lib/elliptic/ec/index.js
wanderer: oh right! taken care of

lib/elliptic/ec/index.js
wanderer: gotcha!

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,39 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+
+  // 1.1. Let x = r + jn.
+  r = this.curve.pointFromX(isYOdd, r);
+  var eNeg = e.neg().mod(n);
+
+  // 1.6.1 Compute Q = r^-1 (sR -  eG)
+  //               Q = r^-1 (sR + -eG)
+  var rInv = signature.r.invm(n);
+
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
```
indutny: Just `return`? ;)

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,39 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
 
   return p.getX().mod(this.n).cmp(r) === 0;
 };
+
+EC.prototype.recoverPubKey = function(msg, signature, j) {
+  assert(3 & j === j, 'The recovery param is more than two bits');
```
indutny: `(3 & j) === j`, current code is doing `3 & (j === j)` ;)

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,39 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  var eNeg = e.neg().mod(n);
+
+  // 1.6.1 Compute Q = r^-1 (sR -  eG)
+  //               Q = r^-1 (sR + -eG)
+  var rInv = signature.r.invm(n);
+  var Q = r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+  return Q;
+};
+
+EC.prototype.calcPubKeyRecoveryParam = function(e, signature, Q, enc) {
```
indutny: What do you think about naming it `getKeyRecoveryParam()`?

lib/elliptic/ec/index.js
wanderer: done

lib/elliptic/ec/index.js
wanderer: oh wow i totally missed that

lib/elliptic/ec/index.js
wanderer: sure done!

indutny:
LGTM, I'll work on few test cases to cover all possibilities of `j`.

indutny:
Thank you!

indutny:
Gosh, can't really find anything for `j > 1`... Do you happen to have any examples?

lib/elliptic/ec/index.js
```diff
@@ -164,3 +164,39 @@ EC.prototype.verify = function verify(msg, signature, key, enc) {
...
+  r = this.curve.pointFromX(isYOdd, r);
+  var eNeg = e.neg().mod(n);
+
+  // 1.6.1 Compute Q = r^-1 (sR -  eG)
+  //               Q = r^-1 (sR + -eG)
+  var rInv = signature.r.invm(n);
+  return r.mul(s).add(this.g.mul(eNeg)).mul(rInv);
+};
+
+EC.prototype.getKeyRecoveryParam = function(e, signature, Q, enc) {
```
indutny: Let me think if I can figure a way to make it simpler... I guess it could be actually returned by the `.sign()` method.

indutny: What do you think about:

```patch
commit 9aa6c6ae603800f76722c5eda4b0cfc0870ffad0
Author: Fedor Indutny <fedor@indutny.com>
Date:   Thu Jun 11 01:37:42 2015 +0200

    ecdsa: return recoveryParam when signing

diff --git a/lib/elliptic/ec/index.js b/lib/elliptic/ec/index.js
index 833e04b..df9f0ea 100644
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -124,7 +124,8 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (kp.isInfinity())
       continue;
 
-    var r = kp.getX().mod(this.n);
+    var kpX = kp.getX();
+    var r = kpX.mod(this.n);
     if (r.cmpn(0) === 0)
       continue;
 
@@ -136,7 +137,10 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (options.canonical && s.cmp(this.nh) > 0)
       s = this.n.sub(s);
 
-    return new Signature({ r: r, s: s });
+    var recoveryParam = (kp.getY().isOdd() ? 1 : 0) |
+                        (kpX.cmp(r) !== 0 ? 2 : 0);
+
+    return new Signature({ r: r, s: s, recoveryParam: recoveryParam });
   } while (true);
 };
 
@@ -192,6 +196,9 @@ EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
 
 EC.prototype.getKeyRecoveryParam = function(e, signature, Q, enc) {
   signature = new Signature(signature, enc);
+  if (signature.recoveryParam !== null)
+    return signature.recoveryParam;
+
   for (var i = 0; i < 4; i++) {
     var Qprime = this.recoverPubKey(e, signature, i);
 
diff --git a/lib/elliptic/ec/signature.js b/lib/elliptic/ec/signature.js
index 08cc84f..0e3da5b 100644
--- a/lib/elliptic/ec/signature.js
+++ b/lib/elliptic/ec/signature.js
@@ -16,6 +16,10 @@ function Signature(options, enc) {
   assert(options.r && options.s, 'Signature without r or s');
   this.r = new bn(options.r, 16);
   this.s = new bn(options.s, 16);
+  if (options.recoveryParam !== null)
+    this.recoveryParam = options.recoveryParam;
+  else
+    this.recoveryParam = null;
 }
 module.exports = Signature;
 
@@ -43,6 +47,7 @@ Signature.prototype._importDER = function _importDER(data, enc) {
 
   this.r = new bn(data.slice(4, 4 + rlen));
   this.s = new bn(data.slice(4 + rlen + 2, 4 + rlen + 2 + slen));
+  this.recoveryParam = null;
 
   return true;
 };
```

If it LGTY, I'll land it with your changes ;)

wanderer: noce! thanks

wanderer:
>Gosh, can't really find anything for j > 1... Do you happen to have any examples?
I didn't the time ATM to follow up on this, but there is a test in bitcoin's repo.
```
<null_radix> does any have an example secp256k1 sig that has a recovery id of 2?
<sipa> it's extremely unlikely to occur, but it is possible
<sipa> and they can be constructed using key recovery
<null_radix> thanks
<sipa> libsecp256k1 actually has unit tests that do this
<null_radix> im checking it out
<null_radix> great, i think i found it https://github.com/bitcoin/secp256k1/blob/master/src/tests.c#L1723
```

indutny:
@wanderer thanks for checking. Unfortunately they don't have example of it either, I think it is very unlikely event when it can actually be >= 2. Like one to 2^128.

Landing!

indutny:
Landed and released in 3.1.0 . Thank you!

sublimator:
Woo!

dcousens:
@indutny examples of `i > 1` can be found [here](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/ecdsa.js#L104-L115)

indutny:
@dcousens I was searching for the example of signature that might be decoded with `i > 1`

indutny:
Thanks, though!

wanderer:
:+1:

sublimator:
Is there any canonical form or standard for byte encoding signatures that include the recovery factor?

ryanxcharles:
Bitcoin has a standard called "compact signatures" that includes the recovery factor:
* https://github.com/ryanxcharles/fullnode/blob/dfd22ea765a386ea5a341ac95ee78370c31b70f3/lib/sig.js#L280
* https://github.com/ryanxcharles/fullnode/blob/dfd22ea765a386ea5a341ac95ee78370c31b70f3/lib/sig.js#L78

sublimator:
Thanks for the info :)

Issue #38, opened 2015-05-13
Test cases don't test sign and verify


clark800:
Test cases only test that signatures verify, which would be trivially satisfied if the verify function always returned true.

https://github.com/indutny/elliptic/blob/c8d7cf551fdf2ce3ecc5264b29084244ae6aa2b2/test/ecdsa-test.js#L64

To actually test signing, it is necessary to test against actual test vectors and negative cases. The official test vectors for ed25519 can be found here: http://ed25519.cr.yp.to/software.html (see sign.input).

indutny:
This is absolutely right! Thanks for pointing it out!

Will you be interested in helping me with this?

sublimator:
To get the test vectors to pass you'll need to also implement EdDSA, which has a different signature scheme etc. 

See:
http://blog.josefsson.org/2015/03/04/eddsa-and-ed25519-goes-to-ietf/
https://tools.ietf.org/html/draft-josefsson-eddsa-ed25519-03#section-5.6

Aside from the e25519 curve using EC)DSA instead of EDDSA, I also had trouble with scalar multiplication. I tried using [scalarbase](https://github.com/dchest/tweetnacl-js/blob/master/nacl.js#L697-L718) from nacl to compute the public key from a 32 byte secret key but with `elliptic.curves.ed25519.g.mul(x)` I could not get anything that corresponded to the 32 bytes encoded public point. 

I'm very much a n00b with this stuff, so I could be doing something wrong. Anyway, I hacked the following into nacl (based off [crypto_sign_keypair] (https://github.com/dchest/tweetnacl-js/blob/master/nacl.js#L706) )  which took the same `x` value I passed to `ed25519.g.mul(x)` but no bueno.

```js
/**
* @param {Array} a - first 32 bytes of sha512 of 256 bit `secret key`
*                    with certain bits already set/cleared via:
*                         a[0] &= 248;
*                         a[31] &= 127;
*                         a[31] |= 64;
*
* @return {Array} public key point
*/
function computePublicKey(a) {
  var p = [gf(), gf(), gf(), gf()];
  var pk = new Uint8Array(32);
  scalarbase(p, a);
  pack(pk, p);
  var pkk = new Array(32);
  for (var i = 0; i < pk.length; i++) {
    pkk[i] = pk[i];
  };
  return pkk;
}
```

[...]

PR #40, opened 2015-06-01
[WIP] Add EdDSA / Ed25519 class and tests

sublimator:
Work in progress. Up for comment.

#38


indutny:
Going to fix the `_getDoubles` thing first, before reviewing this. Thanks for pointing out!

indutny:
Ok, it is fixed in 3.0.4 . Please rebase your PR ;)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,162 @@
...
+
+EDDSA.prototype.encodeInt = function(num) {
+  var bytes = num.toArray();
+  while (bytes.length < this.encBytes)
+    bytes.unshift(0);
+  return bytes.reverse();
+};
+
+EDDSA.prototype.decodeInt = function(bytes) {
+  return new bn(bytes, 'le');
```
sublimator: todo

sublimator:
Man, you [kick ass for the lord!](https://www.youtube.com/watch?v=VYvqOfrs6CA)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,162 @@
...
+
+function EDDSA(curve) {
+  if (!(this instanceof EDDSA))
+    return new EDDSA(curve);
+
+  assert(typeof curve === 'string');
+  assert(curve === 'ed25519', 'only tested with ed25519 so far');
+  var curve = this.curve = elliptic.curves[curve].curve;
+  this.g = curve.g;
+  // Later we do a G.mul(fiveHundredAnd12BitNumber) so we need more doubles.
```
indutny: Guess this is no longer required.

sublimator: Is is faster if we use `_fixedNafMul` somehow ?

sublimator: By very quick crude timing of `npm test`, it does seem be a little bit faster ( ~12% ) when using `_fixedNafMul`

indutny: You may want to do following thing below: `this.g.mul(r.mod(this.curve.n))`

sublimator: Ah right, that should speed things up yeah?

sublimator: And not require 512b scalar ?

sublimator: Any place I do point * scalar, I can use mod(curve.n) first ?

indutny: Yeah, it is the order of group. It cycles back after `curve.n` anyway.

sublimator: Ok, I did that the 2 places I saw it.  

We are checking the bitLength anyway when precomputed, could it just mod(order) transparently ?

sublimator:
Lemme know any changes I should make, also when I should squash this down.

sublimator:
Would you like a keypair class, and genKeyPair() etc API replicated for consistency?

indutny:
@sublimator changes are looking very good. The only question that I have: do you think it will be better to have it as a part of `elliptic`? Maybe it will be simpler to have it as a separate module?

sublimator:
I think we were hoping to be contributing to something where
maintenance/review burden is shared. I get the feeling people will be using
eddsa/ed25519 more in the future. It's probably worth merging? What did you
have in mind when you said another module?

indutny:
I meant [eddsa][https://www.npmjs.com/package/eddsa] npm package. There are several pros in this:

* Separate release cycle
* Smaller size of elliptic.js distribution

I'm absolutely fine reviewing things happening there, can't say that I have much time to maintain one more project, though :(

sublimator:
Will chat with the powers that be :)

clark800:
I'd prefer not to start a new package because the goal here is wider exposure and a separate package reduces discoverability.

With just 158 lines of distribution code, I'd imagine this pull request would have a fairly minimal impact on distribution size.

Also, Wikipedia says "Ed25519 is a specific implementation of EdDSA", so does it make sense to have support for Ed25519 without support for EdDSA?

If you're worried about maintenance effort, would you be interested in having us help out with maintaining the project?

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+module.exports = EDDSA;
+
+/**
+* @param {Array} message - message bytes
+* @param {Array|String} seed - secret seed bytes
+* @returns {bytes} - signature
+*/
+EDDSA.prototype.sign = function(message, secret) {
```
indutny: May I ask you to avoid anonymous functions? `function sign(message, secret) {`... ;)

indutny:
Ok, I guess it makes sense. Though, I'm going to exclude it from the minimal footprint distribution for obvious reasons ;)

lib/elliptic/eddsa/index.js
sublimator: You may :)

indutny: I ask you!

sublimator: I comply

indutny:
May I ask you to squash everything into a single commit?

sublimator:
You may, but I already squashed and pushed, with great -f/orce :)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+
+  var a = hash.slice(0, this.encBytes);
+  a[0] &= 248;
+  a[lastIx] &= 127;
+  a[lastIx] |= 64;
+
+  var aBn = this.decodeRedInt(a);
+  var A = this.g.mul(aBn);
+
+  return {
```
indutny: I would create a class for this.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+         a.getY().cmp(b.getY()) === 0;
+}
+
+/**
+* @param {Array} sig - sig bytes
+* @param {Array} message - message bytes
+* @param {Array|String} A - public key bytes
+* @returns {bytes} - true if
+*/
+EDDSA.prototype.verify = function verify(sig, message, A) {
```
indutny: And accept this class here as an `A`. So it is either `derived` or `A`.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+module.exports = EDDSA;
+
+/**
+* @param {Array} message - message bytes
+* @param {Array|String} seed - secret seed bytes
+* @returns {bytes} - signature
+*/
+EDDSA.prototype.sign = function sign(message, secret) {
```
indutny: Is there any reason to not use `Signature` class that we already have?

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+  message = parseBytes(message);
+  var derived = (Array.isArray(secret) || typeof secret === 'string') ?
+                this.deriveFromSeed(secret) : secret;
+  var r = this.hashToInt(derived.suffix, message);
+  var R = this.encodePoint(this.g.mul(r.mod(this.curve.n)));
+  var s_ = this.hashToInt(R, derived.encoded.A, message).mul(derived.a);
+  var S = r.add(s_).mod(this.curve.n);
+  return R.concat(this.encodeInt(S));
```
indutny: Gosh, that comment belong here ^

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,158 @@
...
+*               Arrays of bytes.
+*/
+EDDSA.prototype.hashToInt = function hashToInt() {
+  var hash = this.hash();
+  for (var i = 0; i < arguments.length; i++)
+    hash.update(arguments[i]);
+  return this.decodeInt(hash.digest());
+};
+
+EDDSA.prototype.deriveFromSeed = function deriveFromSeed(seed) {
```
indutny: When you call it like this, I expect that there is some `.derive()` method that I am not aware of. Any reason to not name it `.derive()`?

lib/elliptic/eddsa/index.js
sublimator: That makes sense, do you want a full KeyPair class etc? I can try and match the api with EC?

It's 1:30am here, and I'm about to sleep. I'm more than happy if you form it as you'd like yourself if you are :) Otherwise tomorrow.

sublimator: Ditto for verify taking derived bundle as well as A as point or encoded, or whatever.

indutny: Yeah, though, you may need to introduce a new keypair class, since the public key derivation is different.

lib/elliptic/eddsa/index.js
sublimator: I wasn't super content with this either. I think the FromSeed part bubble
up from the nacl api: `nacl.sign.keypair.fromSeed` or something.

derive works :)

lib/elliptic/eddsa/index.js
sublimator: Yeah, I didn't think there was much to be gained trying to extend from EC/*
eh

lib/elliptic/eddsa/index.js
sublimator: EdDSA doesn't use DER encoding for 1

lib/elliptic.js
```diff
@@ -11,3 +11,4 @@ elliptic.curves = require('./elliptic/curves');
 
 // Protocols
 elliptic.ec = require('./elliptic/ec');
+elliptic.eddsa = require('./elliptic/eddsa');
```
sublimator: `ed` vs `eddsa` ?

indutny: I would name it `eddsa`.

sublimator: :)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,185 @@
...
+      pubBytes = this.encodePoint(pub);
+    } else {
+      pubBytes = parseBytes(pub);
+      pub = this.decodePoint(pubBytes);
+    }
+    return new KeyPair(this, { pub: pub, pubBytes: pubBytes });
+  }
+};
+
+EDDSA.prototype.keyFromSecret = function keyFromSecret(secret) {
```
sublimator: I don't want to use private, perhaps `seed` would be better? because as you can see the secret/seed is hashed, and the first half is used, bar a little bit twiddling as the `priv`, the scalar multiplied by the base point to find the public key. The remaining half is used as a prefix to the message hash. 

`keyFromPrivate` would be more consistent with `EC` but ..... ? do you like secret or seed ? or you're OK with private?

indutny: This name makes sense.

lib/elliptic/eddsa/key.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var elliptic = require('../../elliptic');
+var utils = elliptic.utils;
+var assert = utils.assert;
+
+function KeyPair(eddsa, params) {
+  this.eddsa = eddsa;
+  this.pubBytes = params.pubBytes;
```
indutny: I think it should be capable of deriving the public side itself.

sublimator:
There' s definitely not much that can be done extending from EC/signature.js somehow, as EdDSA has a point for R, and a scalar for S, nor does it use DER encoding. I'm not sure what adding a new Signature class would bring as they are always 64 bytes. Perhaps maybe for benchmarking sign/verify, to avoid the point decoding :) 

?

lib/elliptic/eddsa/key.js
sublimator: It can certainly be done, but then you'd need to get at everything through
accessors. Not necessarily a bad thing of course ...

Note that the public key bytes are hashed when signing, implying a Point
along the way, so I don't think there's much to be gained, by doing it
lazily.

But I can make it use accessors and do lazy member computation if you want.

indutny: Yes, accessors please! :) And lazy computation to, I think this is how it is done for the ECDSA keypair now.

indutny:
@sublimator why not encode it in DER? How would you store it otherwise?

indutny:
@sublimator also, I didn't mean to always encode/decode it. It should just have APIs for accepting both DER and JS object, using the signature class.

sublimator:
They are always of a consistent length, 32 bytes R, encoded point, and 32
bytes S, encoded int. It's just stored as a concatenation of the two, as a
byte array.

sublimator:
EdDSA never uses DER though ?

I can add it though if you want.

indutny:
Ooooh, you are right. Anyway, this should live in a separate class (I mean the code for splitting the 64 bytes R+S into the components).

sublimator:
Ok :)  I didn't bother yet cause it's basically R.concat(S), and
sig.slice(0, 32), sig.slice(32) to destructure it.

But it would be a bit cleaner and more consistent huh :)

I'm on it::
KeyPair to use accessors + lazy computation
Add Signature class

sublimator:
And about the secret/seed/private terminology  ?

indutny:
I'd go for `secret`

sublimator:
Cool, agreed :)

indutny:
Thank you!

sublimator:
First stab at lazy derivations up ^

lib/elliptic/eddsa/key.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+var elliptic = require('../../elliptic');
+var utils = elliptic.utils;
+var assert = utils.assert;
+
+function lazyComputed(obj, name, computer) {
```
indutny: Let's put it to utils?

lib/elliptic/eddsa/key.js
```diff
@@ -0,0 +1,92 @@
...
+  var key = '_' + name;
+  obj.prototype[name] = function() {
```
indutny: I suggest giving this function a name.

test/ed25519-test.js
```diff
@@ -0,0 +1,137 @@
...
+    var derived = ed25519.keyFromSecret(secret);
+    var sig = toHex(ed25519.sign(msg, derived));
+
+    var R = "8F1B9A7FDB22BCD2C15D4695B1CE2B063CBFAEC9B00BE360427BAC9533943F6C";
+    var S = "5F0B380FD7F2E43B70AB2FA29F6C6E3FFC1012710E174786814012324BF19B0C";
+
+    assert.equal(sig.slice(0, 64), R);
+    assert.equal(sig.slice(64), S);
+
+    assert(ed25519.verify(msg, sig, derived.pub()));
```
indutny: `derived.pub()` should not be required here. Let's keep it simple :)

sublimator: I don't think it is required ... just the test ...

sublimator: Is there any javascript hack to give the lazyCompute factory accessor
returned a name, that would be useful for stack traces etc ?

sublimator: Hrmm, Signature would need an eddsa instance to decodePoints, if the
constructor took say the signature bytes in some form to decode r/s.

indutny: I'm not aware of such hack.

indutny: That's fine, it is created internally anyway.

sublimator: just not quiteness as niceness :)

sublimator:
@indutny 

Can you take another look?  

Thanks.

lib/elliptic/curve/edwards.js
```diff
@@ -61,7 +61,37 @@ EdwardsCurve.prototype.pointFromX = function pointFromX(odd, x) {
   if (odd && !isOdd || !odd && isOdd)
     y = y.redNeg();
 
-  return this.point(x, y, curve.one);
+  return this.point(x, y, this.one);
```
indutny: Ouch, a bug fix!

lib/elliptic/curve/edwards.js
```diff
@@ -61,7 +61,37 @@ EdwardsCurve.prototype.pointFromX = function pointFromX(odd, x) {
...
+};
+
+EdwardsCurve.prototype.pointFromY = function pointFromY(odd, y) {
```
indutny: Maybe reverse the arguments: `y, odd`?

lib/elliptic/curve/edwards.js
```diff
@@ -61,7 +61,37 @@ EdwardsCurve.prototype.pointFromX = function pointFromX(odd, x) {
...
+  var y2 = y.redSqr();
+  var lhs = y2.redSub(curve.one);
+  var rhs = y2.redMul(curve.d).redAdd(curve.one);
+  var x2 = lhs.redMul(rhs.redInvm());
+
+  if (x2.cmpn(0) === 0) {
+    if (odd)
+      throw new Error('invalid point');
+    else
+      return curve.point(new bn(0).toRed(curve.red), y);
```
indutny: I guess zero could be potentially cached in the curve itself...

indutny: It already is!

lib/elliptic/curve/edwards.js
```diff
@@ -61,7 +61,37 @@ EdwardsCurve.prototype.pointFromX = function pointFromX(odd, x) {
...
+  }
+
+  var x = x2.redSqrt();
+  if (x.redSqr().redSub(x2).cmpn(0) !== 0)
+    throw new Error('invalid point');
+
+  if (x.isOdd() !== odd)
+    x = x.redNeg();
+
+  return curve.point(x, y, curve.one);
```
indutny: No need to pass `curve.one`, it is handled in `curve.point()` automatically.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+
+var bn = require('bn.js');
+var hash = require('hash.js');
+var elliptic = require('../../elliptic');
+var utils = elliptic.utils;
+var assert = utils.assert;
+var KeyPair = require('./key');
+var Signature = require('./signature');
+
+function parseBytes(bytes) {
```
indutny: Move it to utils?

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+                                     bytes;
+}
+
+function EDDSA(curve) {
+  if (!(this instanceof EDDSA))
+    return new EDDSA(curve);
+
+  assert(typeof curve === 'string');
+  assert(curve === 'ed25519', 'only tested with ed25519 so far');
+  var curve = this.curve = elliptic.curves[curve].curve;
```
indutny: Let's split it into two lines:

```js
this.curve = ...
curve = this.curve
```

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  return typeof bytes === 'string' ? utils.toArray(bytes, 'hex') :
+                                     bytes;
+}
+
+function EDDSA(curve) {
+  if (!(this instanceof EDDSA))
+    return new EDDSA(curve);
+
+  assert(typeof curve === 'string');
+  assert(curve === 'ed25519', 'only tested with ed25519 so far');
```
indutny: Any reason for this? Or just precaution?

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  var r = this.hashToIntModN(key.messagePrefix(), message);
+  var R = this.g.mul(r);
+  var Rencoded = this.encodePoint(R);
+  var s_ = this.hashToIntModN(Rencoded, key.pubBytes(), message)
+               .mul(key.priv());
+  var S = r.add(s_).mod(this.curve.n);
+  return this.makeSignature({ R: R, S: S, Rencoded: Rencoded });
+};
+
+function pointEq(a, b) {
```
indutny: This begs to be a `Point` method ;)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  var SG = this.g.mul(sig.S());
+  var RplusAh = sig.R().add(key.pub().mul(h));
+  return pointEq(RplusAh, SG);
+};
+
+/**
+* @returns {bn} little-endian intepretation of hash of one or more
+*               Arrays of bytes.
+*/
+EDDSA.prototype.hashToIntModN = function hashToIntModN() {
```
indutny: To utils it goes!

indutny: And gets an argument `n`!

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  var hash = this.hash();
+  for (var i = 0; i < arguments.length; i++)
+    hash.update(arguments[i]);
+  return this.decodeInt(hash.digest()).mod(this.curve.n);
+};
+
+EDDSA.prototype.keyFromPublic = function keyFromPublic(pub) {
+  if (pub instanceof KeyPair) {
+    return pub;
+  } else {
```
indutny: Let's skip creating the block and put this stuff right after the early return on the previous line.

lib/elliptic/curve/edwards.js
sublimator: I just copied the curve.one stuff and odd as first arg as a nod to the
existing method. Will fix em both?

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+    var pubIsPoint = this.isPoint(pub);
+    var pubBytes = !pubIsPoint ? parseBytes(pub) : undefined;
+    pub = pubIsPoint ? pub : undefined;
+    return new KeyPair(this, { pub: pub, pubBytes: pubBytes });
+  }
+};
+
+EDDSA.prototype.keyFromSecret = function keyFromSecret(secret) {
+  if (secret instanceof KeyPair) {
```
indutny: Superfluous braces.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+};
+
+EDDSA.prototype.encodePoint = function encodePoint(point) {
+  var enc = this.encodeInt(point.getY());
+  enc[this.encBytes - 1] |= point.getX().isOdd() ? 0x80 : 0;
+  return enc;
+};
+
+EDDSA.prototype.makeSignature = function makeSignature(sig) {
+  if (sig instanceof Signature) return sig;
```
indutny: I'd put `return` on the next line.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  if (typeof sig !== 'object') sig = parseBytes(sig);
```
indutny: Ditto.

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  var lastIx = this.encBytes - 1;
+
+  var normed = bytes.slice(0, lastIx).concat(bytes[lastIx] & ~0x80);
+  var odd = Boolean(bytes[lastIx] & 0x80);
+
+  var y = this.decodeRedInt(normed);
+  return curve.pointFromY(odd, y);
+};
+
+EDDSA.prototype.encodeInt = function encodeInt(num) {
```
indutny: To utils!

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  var bytes = num.toArray();
+  while (bytes.length < this.encBytes)
+    bytes.unshift(0);
+  return bytes.reverse();
+};
+
+EDDSA.prototype.decodeInt = function decodeInt(bytes) {
```
indutny: Utils? ;)

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,143 @@
...
+  return new bn(parseBytes(bytes), 'le');
+};
+
+EDDSA.prototype.decodeRedInt = function decodeRedInt(bytes) {
```
indutny: Ditto.

lib/elliptic/eddsa/key.js
```diff
@@ -0,0 +1,91 @@
...
+*
+* Any optional params not passed in, will be computed lazily (when possible)
+*/
+function KeyPair(eddsa, params) {
+  var self = this;
+  this.eddsa = eddsa;
+
+  // Set all params passed in as private members, others will be computed lazily
+  // where possible.
+  Object.keys(params).forEach(function(k) {
```
indutny: Yikes, this plays badly with hidden classes. Let's always define all params and put the manually, shouldn't be that much of them.

lib/elliptic/eddsa/signature.js
```diff
@@ -0,0 +1,58 @@
...
+
+  if (Array.isArray(sig))
+    sig = {
+      R: sig.slice(0, eddsa.encBytes),
+      S: sig.slice(eddsa.encBytes)
+    };
+
+  assert(sig.R && sig.S, 'Signature without R or S');
+
+  if (eddsa.isPoint(sig.R)) this._R = sig.R;
```
indutny: Body of if should be on next line.

lib/elliptic/eddsa/signature.js
```diff
@@ -0,0 +1,58 @@
...
+  if (sig.S instanceof bn) this._S = sig.S;
```
indutny: Ditto.

lib/elliptic/eddsa/signature.js
```diff
@@ -0,0 +1,58 @@
...
+* @param {Array<Bytes>|Object} sig -
+* @param {Array<Bytes>|Point} [sig.R] - R point as Point or bytes
+* @param {Array<Bytes>|bn} [sig.S] - S scalar as bn or bytes
+* @param {Array<Bytes>} [sig.Rencoded] - R point encoded
+* @param {Array<Bytes>} [sig.Sencoded] - S scalar encoded
+*/
+function Signature(eddsa, sig) {
+  this.eddsa = eddsa;
+
+  if (Array.isArray(sig))
```
indutny: Needs braces around the body.

lib/elliptic/utils.js
```diff
@@ -148,3 +148,12 @@ function getJSF(k1, k2) {
   return jsf;
 }
 utils.getJSF = getJSF;
+
+function lazyComputed(obj, name, computer) {
```
indutny: Good! Here is the right place for this ;)

test/ed25519-test.js
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  function testFactory(i) {
+    it('can compute correct a and A for secret: ' + i, function() {
+      var test = derivations[i];
+      var secret = utils.toArray(test.secret_hex, 'hex');
+      var key = ed25519.keyFromSecret(secret);
+      assert.equal(toHex(key.privBytes()), test.a_hex);
+      var xRecovered = toHex(ed25519.encodeInt(
+                              ed25519.decodePoint(key.pubBytes()).getX()));
```
indutny: Borked alignement ;)

indutny: Btw, why so much magic? Isn't it just possible to do `.getX().toString('hex', 2)`?

lib/elliptic/curve/edwards.js
indutny: Sounds good.

sublimator: I didn't really want the point/int codecs on the EDDSA class, but
ultimately the curve encoding size byte lengh may differ. I guess the
encsize could be passed as a parameter.

Stuck in the rain atm.

test/ed25519-test.js
sublimator: The ints are little endian. I'm guessing the toString('hex') won't work.

lib/elliptic/eddsa/key.js
sublimator: Yes theres only really 3. Will make it explicit

lib/elliptic/eddsa/index.js
sublimator: Yeah, just to be clear, and only taking string of ed25519 for the moment
easy wat of checking the only tested curve is used.

lib/elliptic/eddsa/index.js
sublimator: The hash function used, needs to be double bit length of order (or prime I
can't remember). The impls of some of these methods u want in utils will
need parameterising / renaming. Sha512? Little endian? Certain encoding
length. I'd still like to have the convenience methods without needing to
parameterise everywhere. Codec class? Lol

indutny: `hash` and `n` for this function ;)

indutny:
Is it ready for review? ;)

sublimator:
For another round at least :)

sublimator:
I wouldn't have extracted the utils until they were needed by something
else,  personally. And some of it could be put in bn.js, for example padded
little endian toArray()

sublimator:
Also, I slapped the Point eq method into the edwards impl. I see the short
curve point has an implementation. Not sure if it can be pushed up into the
base class and shared?

sublimator:
@indutny 

I rebased onto 3.1.0. Can you take another look?  I enabled push access for you on sublimator/elliptic if you want to just make any changes you want.

sublimator:
bump :)

sublimator:
bump :) 

@indutny

lib/elliptic/eddsa/index.js
```diff
@@ -0,0 +1,103 @@
...
+
+  var curve = elliptic.curves[curve].curve;
+  this.curve = curve;
+  this.g = curve.g;
+  this.g.precompute(curve.n.bitLength() + 1);
+
+  this.pointClass = curve.point().constructor;
+  this.encodingLength = Math.ceil(curve.n.bitLength() / 8);
+  this.hash = hash.sha512;
+  this.hashToIntModOrder = utils.hashToLEIntModN.bind(null, this.hash, curve.n);
```
indutny: I'm against using `.bind()`, for performance reasons. Let' maske this one a method that will call the `utils.*` one.

lib/elliptic/eddsa/signature.js
```diff
@@ -0,0 +1,65 @@
...
+* @param {Array<Bytes>} [sig.Rencoded] - R point encoded
+* @param {Array<Bytes>} [sig.Sencoded] - S scalar encoded
+*/
+function Signature(eddsa, sig) {
+  this.eddsa = eddsa;
+
+  if (typeof sig !== 'object')
+    sig = parseBytes(sig);
+
+  if (Array.isArray(sig))
```
indutny: Needs `{` `}`

lib/elliptic/utils.js
```diff
@@ -37,7 +38,6 @@ function toArray(msg, enc) {
   }
   return res;
 }
-utils.toArray = toArray;
```
indutny: Why? :sob:

test/ed25519-test.js
```diff
@@ -0,0 +1,138 @@
...
+var elliptic = require('../');
+var utils = elliptic.utils;
+var toArray = elliptic.utils.toArray;
+var eddsa = elliptic.eddsa;
+
+function toHex(arr) {
+  return elliptic.utils.toHex(arr).toUpperCase();
+}
+
+var MAX_PROGRAMMATIC = process.env.CI ? Infinity : 50;
```
indutny: Is it to hard for CI? ;)

lib/elliptic/eddsa/index.js
indutny: Let's give it a shorter name: `hashToRedInt` or `...ModInt`, order doesn't really apply to it as it is too generic.

lib/elliptic/utils.js
```diff
@@ -147,4 +144,66 @@ function getJSF(k1, k2) {
...
+           this[key] = computer.apply(this, arguments);
+  };
+}
+
+function parseBytes(bytes) {
+  return typeof bytes === 'string' ? utils.toArray(bytes, 'hex') :
+                                     bytes;
+}
+
+function intFromLE(bytes) {
```
indutny: I'll work on doing in `bn.js`.

lib/elliptic/eddsa/index.js
sublimator: I didn't want to put anything in utils that assumed LE without being clear about it. On that matter I'd have preferred to just maybe just leave them in this file, in a `staging area` or something :)

Anyway, which are you talking about?  `hashToLEIntModN`? or `hashToIntModOrder` ?  The latter binds the curve order as the n.

lib/elliptic/utils.js
sublimator: hahaha ?  You want the exports always after the definition?

test/ed25519-test.js
sublimator: This is to allow the CI to run all the test vectors, but only 50 when running locally. Because it can be a bit of a pain to wait for them all.

lib/elliptic/utils.js
sublimator: That would be great! I opened up some issues re: toArray and honouring the `le` when passing in array.

lib/elliptic/eddsa/signature.js
sublimator: What's your rule for when requiring `{}`? Multiple lines, not just multiple expressions?

lib/elliptic/eddsa/index.js
sublimator: > order doesn't really apply to it as it is too generic.
> Let's give it a shorter name:

I'm still confused which function you're talking about.

indutny: Oooh, they both have huge names. I'm sorry to say it, but let's put it back into EDDSA file, and give it a short name.

lib/elliptic/utils.js
```diff
@@ -147,4 +144,66 @@ function getJSF(k1, k2) {
...
+  var normed = bytes.slice(0, lastIx).concat(bytes[lastIx] & ~0x80);
+  var xIsOdd = Boolean(bytes[lastIx] & 0x80);
+
+  var y = intFromLE(normed);
+  return curve.pointFromY(y, xIsOdd);
+}
+
+function hashToLEIntModN(hashFactory, n) {
+  var hash = hashFactory();
+  Array.prototype.slice.call(arguments, 2).forEach(hash.update.bind(hash));
```
indutny: No `.bind()`, please ;) Also, what do you think about manually going through the arguments in a for loop starting from 2? This should be much faster AFAIK.

lib/elliptic/utils.js
indutny: Yeah, I can't even check if you exported everything without checking every name.

test/ed25519-test.js
indutny: Oook, guess it makes sense...

lib/elliptic/eddsa/signature.js
indutny: Yeah, single line doesn't need it. Multi-line needs it. That's it :)

lib/elliptic/eddsa/index.js
sublimator: You want to move back just that particular function or all the new utils ?

indutny: Just this one. :)

lib/elliptic/utils.js
indutny: Please update bn.js to 2.1.0 here and get all the awesomeness out of it :)

lib/elliptic/eddsa/index.js
sublimator: How do you feel about this:
```javascript
EDDSA.prototype.hashInt = function hashInt() {
  var hash = this.hash();
  for (var i = 0; i < arguments.length; i++) {
    hash.update(arguments[i]);
  }
  return utils.intFromLE(hash.digest()).mod(this.curve.n);
};
```

?

lib/elliptic/utils.js
sublimator: Vill do!

lib/elliptic/eddsa/index.js
indutny: Great!

sublimator: ok, pushed again (while you're here :)

I think I addressed most of your comments

checking

lib/elliptic/utils.js
```diff
@@ -147,4 +148,54 @@ function getJSF(k1, k2) {
...
+
+function intToLE(num, padTo) {
+  var bytes = num.toArray('le');
+  while (bytes.length < padTo)
+    bytes.push(0);
+  return bytes;
+}
+exports.intToLE = intToLE;
+
+function pointToLEYoddX(point, length) {
```
sublimator: Are you sure you want these here? It might be nice to have encode methods on Point classes at some point? My feeling about utils is, if you export them, somebody might gonna use em.

lib/elliptic/utils.js
sublimator: Oops, just noticed I exported on `exports` not `utils` as before. I pushed another commit.

indutny:
LGTM. Let's land this thing! Do you promise to take care of it? ;)

sublimator:
Haha sure, we want to use this and the whole elliptic family :) I just
ducked off for breakfast. Will rebase/squash and push in maybe 30 minutes.

sublimator:
"In sickness and in health ... " mumble mumble

indutny:
I will squash it myself, no worries. I'll remember your promise ;)

indutny:
Landed in c0690b36be043ee73c1780ae4b7df48632b11cf9, thank you!

sublimator:
Yay! Thanks :)

clark800:
Congratulations! Great work!

sublimator:
w00t :) we r inching closer to ed25519 support

Issue #42, opened 2015-06-13
Point encoding 


sublimator:
```javascript
function compressedPoint(p) {
  const x = p.getX().toArray();
  const y = p.getY();
  while (x.length < 32) x.unshift(0)
  return [y.isEven() ? 0x02 : 0x03].concat(x);
}
```

Currently for implementing some seed -> secret routines (the algorithm is set in stone) which involves hashing byte representations of points, I needed a little utility function. I note the only code for this currently in elliptic is in the EC/KeyPair#getPublic method and not immediately useful.

Maybe some Point.encode(encoding) / curve.decodePoint(encoding) methods? Shrug?

indutny:
I guess this sounds reasonable to me. Feel free to open PR for this.

indutny:
Though, it will need to use byte length of `curve.p` instead of `32`.

sublimator:
Of course :) Was just example from se application cose

sublimator:
Code. Damn mobile :)

PR #48, opened 2015-07-07
[WIP] Move point codec code from ec/key to curve/base

sublimator:
re #42 

This just takes the existing point codec code that was inside the KeyPair class and moves it the BaseCurve / BasePoint classes. It didn't really seem advantageous to go to unnatural lengths to have default implementation and various over-rides. Thoughts?

### DIFF
--- a/lib/elliptic/curve/base.js
+++ b/lib/elliptic/curve/base.js
@@ -2,10 +2,11 @@
 
 var bn = require('bn.js');
 var elliptic = require('../../elliptic');
-
-var getNAF = elliptic.utils.getNAF;
-var getJSF = elliptic.utils.getJSF;
-var assert = elliptic.utils.assert;
+var utils = elliptic.utils;
+var getNAF = utils.getNAF;
+var getJSF = utils.getJSF;
+var assert = utils.assert;
+var intToBE = utils.intToBE;
 
 function BaseCurve(type, conf) {
   this.type = type;
@@ -239,8 +240,49 @@ function BasePoint(curve, type) {
 }
 BaseCurve.BasePoint = BasePoint;
 
+BasePoint.prototype.eq = function eq(/*other*/) {
+  throw new Error('Not implemented');
+};
+
 BasePoint.prototype.validate = function validate() {
   return this.curve.validate(this);
+};
+
+BaseCurve.prototype.decodePoint = function decodePoint(bytes, enc) {
+  bytes = utils.toArray(bytes, enc);
+
+  var len = this.p.byteLength();
+  if (bytes[0] === 0x04 && bytes.length - 1 === 2 * len) {
+    return this.point(bytes.slice(1, 1 + len),
+                      bytes.slice(1 + len, 1 + 2 * len));
+  } else if ((bytes[0] === 0x02 || bytes[0] === 0x03) &&
+              bytes.length - 1 === len) {
+    return this.pointFromX(bytes.slice(1, 1 + len), bytes[0] === 0x03);
+  }
+  throw new Error('Unknown point format');
+};
+
+BasePoint.prototype.encodeCompressed = function encodeCompressed(enc) {
+  return this.encode(true, enc);
+};
+
+BasePoint.prototype._encode = function _encode(compact) {
+  var len = this.curve.p.byteLength();
+  var x = intToBE(this.getX(), len);
+
+  return compact ?
+      [ this.getY().isEven() ? 0x02 : 0x03 ].concat(x) :
+      [ 0x04 ].concat(x, intToBE(this.getY(), len)) ;
+};
+
+BasePoint.prototype.encode = function encode(compact, enc) {
+  // compact is optional argument
+  if (typeof compact === 'string') {
+    enc = compact;
+    compact = null;
+  }
+
+  return utils.encode(this._encode(compact), enc);
 };
 
 BasePoint.prototype.precompute = function precompute(power) {
--- a/lib/elliptic/curve/mont.js
+++ b/lib/elliptic/curve/mont.js
@@ -5,6 +5,9 @@ var bn = require('bn.js');
 var inherits = require('inherits');
 var Base = curve.base;
 
+var elliptic = require('../../elliptic');
+var utils = elliptic.utils;
+
 function MontCurve(conf) {
   Base.call(this, 'mont', conf);
 
@@ -42,6 +45,10 @@ function Point(curve, x, z) {
 }
 inherits(Point, Base.BasePoint);
 
+MontCurve.prototype.decodePoint = function decodePoint(bytes, enc) {
+  return this.point(utils.toArray(bytes, enc), 1);
+};
+
 MontCurve.prototype.point = function point(x, z) {
   return new Point(this, x, z);
 };
@@ -52,6 +59,10 @@ MontCurve.prototype.pointFromJSON = function pointFromJSON(obj) {
 
 Point.prototype.precompute = function precompute() {
   // No-op
+};
+
+Point.prototype._encode = function _encode(/*compact ignored*/) {
+  return utils.intToBE(this.getX(), this.curve.p.byteLength());
 };
 
 Point.fromJSON = function fromJSON(curve, obj) {
@@ -145,6 +156,10 @@ Point.prototype.mul = function mul(k) {
 
 Point.prototype.mulAdd = function mulAdd() {
   throw new Error('Not supported on Montgomery curve');
+};
+
+Point.prototype.eq = function eq(other) {
+  return this.getX().cmp(other.getX()) === 0;
 };
 
 Point.prototype.normalize = function normalize() {
--- a/test/curve-test.js
+++ b/test/curve-test.js
@@ -162,3 +162,86 @@ describe('Curve', function() {
     assert(p.eq(neg2));
   });
 });
+
+describe('Point codec', function () {
+  function makeShortTest(definition) {
+    var curve = elliptic.curves.secp256k1.curve;
+    return function() {
+      var co = definition.coordinates;
+      var p = curve.point(co.x, co.y);
+
+      // Encodes as expected
+      assert.equal(p.encode('hex'), definition.encoded);
+      assert.equal(p.encodeCompressed('hex'), definition.compactEncoded);
+
+      // Decodes as expected
+      assert(curve.decodePoint(definition.encoded, 'hex').eq(p));
+      assert(curve.decodePoint(definition.compactEncoded, 'hex').eq(p));
+    }
+  }
+
+  function makeMontTest(definition) {
+    var curve = elliptic.curves.curve25519.curve;
+
+    return function() {
+      var co = definition.coordinates;
+      var p = curve.point(co.x, co.z);
+      var encoded = p.encode('hex');
+      var decoded = curve.decodePoint(encoded, 'hex');
+      assert(decoded.eq(p));
+      assert.equal(encoded, definition.encoded);
+    }
+  }
+
+  var shortPointEvenY = {
+    coordinates: {
+      x: "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+      y: "483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8",
+    },
+    compactEncoded:
+      "02" +
+      "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+    encoded:
+      "04" +
+      "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798" +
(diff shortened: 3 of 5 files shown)


sublimator:
Something to consider re: point codec, that ed25519 style (Edwards? ) curves here support both eddsa and ecdsa. Which is the native encoding? 

Currently the `enc` parameter defines the byte encoding, rather than how the point itself is encoded. We could alter the API such that there was some way of controlling the latter.

lib/elliptic/curve/base.js
```diff
@@ -34,7 +34,6 @@ module.exports = BaseCurve;
 BaseCurve.prototype.point = function point() {
   throw new Error('Not implemented');
 };
-
```
indutny: Superfluous change.

indutny:
I'd go with ecdsa native :) LGTM, maybe we should add a tests?

lib/elliptic/curve/base.js
```diff
@@ -243,6 +242,52 @@ BasePoint.prototype.validate = function validate() {
   return this.curve.validate(this);
 };
 
+BaseCurve.prototype.decodePoint = function decodePoint(bytes, enc) {
+  bytes = utils.toArray(bytes, enc);
+
+  if (this.type !== 'mont') {
```
dcousens: Now that we are on the `Curve` object, couldn't you then specialize this per curve to avoid this branch?
May not be worth it for the extra SLOC,  but,  still.

sublimator: Yeah, I considered that. Didn't feel like it would buy much (at least yet)
and easier to review a simple move. You have any strong feelings either
way? I didn't actually try it on for size yet.

sublimator: In any case, will add some tests sometime today :)

dcousens: @sublimator I sent you a PR,  figured I'd see what it ended up looking like.

sublimator: Sweet, on mobile. Will look in 10 :)

lib/elliptic/curve/base.js
```diff
@@ -243,6 +242,48 @@ BasePoint.prototype.validate = function validate() {
...
+  }
+
+  var len = this.curve.p.byteLength();
+  var x = this.getX().toArray();
+
+  for (var i = x.length; i < len; i++)
+    x.unshift(0);
+
+  var res;
+  if (this.curve.type !== 'mont') {
```
sublimator: @dcousens 

For symmetry we'd want to specialise for the Montgomery point encode too? How's that going to look?

lib/elliptic/curve/base.js
sublimator: Will revert

sublimator: Wait, you meant the empty line removed?

lib/elliptic/curve/base.js
sublimator: @dcousens  

By the way I gave you push access to sublimator/elliptic so it's easier to play along

dcousens: Sounds good, will get to this later today :)

sublimator: Cool, I added a few tests for short/mont curve points in the mean time :) 

Confused myself for a bit creating a fixture for mont curves via `short.g.mul(scalar)`. Groan :/

test/curve-test.js
```diff
@@ -162,3 +162,83 @@ describe('Curve', function() {
...
+
+  function makeShortTest(definition) {
+    var curve = shortCurve;
+    return function() {
+      var co = definition.coordinates;
+      var p = curve.point(co.x, co.y);
+
+      // Encodes as expected
+      assert.equal(p.encode(false, 'hex'), definition.encoded);
+      assert.equal(p.encode(true, 'hex'), definition.compactEncoded);
```
sublimator: I'm not in principle a fan of positional boolean args

test/curve-test.js
```diff
@@ -162,3 +162,83 @@ describe('Curve', function() {
...
+
+      // Decodes as expected
+      assert(curve.decodePoint(definition.encoded, 'hex').eq(p));
+      assert(curve.decodePoint(definition.compactEncoded, 'hex').eq(p));
```
sublimator: Because we are using a string, we route to utils.toArray, which interestingly seems to default to assuming each character in the string decodes to a byte. `hex` must explicitly be passed along. I wonder if it would make more sense to default to `hex` for string inputs ?

lib/elliptic/curve/mont.js
```diff
@@ -147,6 +156,10 @@ Point.prototype.mulAdd = function mulAdd() {
   throw new Error('Not supported on Montgomery curve');
 };
 
+Point.prototype.eq = function eq(other) {
```
sublimator: I think I added an `eq` method to the edwards curve, and short already had one. Maybe it would be good to declare an  abstract method using the preexisting convention, similar to:

```javascript
BaseCurve.prototype.point = function point() {
  throw new Error('Not implemented');
};
```
eg.

```javascript
BasePoint.prototype.eq = function eq(other) {
  throw new Error('Not implemented');
};
```

Doesn't seem there's much in common

test/curve-test.js
sublimator: Or in the spirit of `explicit is better`, if there's a string, demand an encoding option passed along.

test/curve-test.js
dcousens: I'm guessing from experience (can't see the code right now) that you are referring to the `compressed` flag in `encode`?

lib/elliptic/curve/mont.js
dcousens: Sounds OK to me, @indutny?

test/curve-test.js
sublimator: Yep, it's `compact` but same diff as `compressed` :)  I just moved it from
the key.getPub() method.

sublimator: > I personally can't stand argument objects as an alternative

I'll count that as a ***strongly*** against any {arguments : 'map'} :)

> encodeCompressed is just as lame.

That doesn't seem so bad to me :) I think I was dropped on my head as a child, and then wrote too much Java though :)

sublimator: This application code is why I bothered to mention it:

```javascript

hasCachedProperty(Secp256k1Pair, 'pubKeyCanonicalBytes', function() {
  return this.key.getPublic(/*compact*/ true, /*enc*/ 'bytes');
});
```

Using comments to remind me wtf the boolean flag meant. Not a big deal. Comments work for me.

lib/elliptic/curve/base.js
sublimator: @dcousens  
You still going to have a play at refactoring this to use specialising methods?

test/curve-test.js
dcousens: Don't get me wrong, `encodeCompressed` is what I would opt for, and have previously when facing this issue.

lib/elliptic/curve/base.js
sublimator: Bump :)

@dcousens @indutny 

I squashed down the progress thus.  Any ideas? I still feel like there's not much to be gained from specialised overrides for Point#encode. Maybe at some point later. I added a delegating Point#encodeCompressed method.

lib/elliptic/curve/mont.js
indutny: OK ;)

sublimator: Bumpy bump :)

dcousens: @sublimator whats up?

lib/elliptic/curve/base.js
```diff
@@ -239,10 +239,60 @@ function BasePoint(curve, type) {
...
+  bytes = utils.toArray(bytes, enc);
+
+  var len = this.p.byteLength();
+  if (bytes[0] === 0x04 && bytes.length - 1 === 2 * len) {
+    return this.point(bytes.slice(1, 1 + len),
+                      bytes.slice(1 + len, 1 + 2 * len));
+  } else if ((bytes[0] === 0x02 || bytes[0] === 0x03) &&
+              bytes.length - 1 === len) {
+    return this.pointFromX(bytes.slice(1, 1 + len), bytes[0] === 0x03);
+  }
```
indutny: What if none of these conditions are met? Should we throw?

lib/elliptic/curve/base.js
```diff
@@ -239,10 +239,60 @@ function BasePoint(curve, type) {
...
+    if (compact) {
+      res = [ this.getY().isEven() ? 0x02 : 0x03 ].concat(x);
+    } else {
+      var y = this.getY().toArray();
+      for (var i = y.length; i < len; i++)
+        y.unshift(0);
+      var res = [ 0x04 ].concat(x, y);
+    }
+  } else {
+    res = x;
```
indutny: This is not going to play well with `decodePoint`, right?

lib/elliptic/curve/mont.js
```diff
@@ -42,6 +45,12 @@ function Point(curve, x, z) {
 }
 inherits(Point, Base.BasePoint);
 
+MontCurve.prototype.decodePoint = function decodePoint(bytes, enc) {
+  bytes = utils.toArray(bytes, enc);
+
+  return this.point(bytes, 1);
```
indutny: Ah, I see.

indutny: Feels odd, let's have the encode function in this file too.

indutny:
One nit, otherwise LGTM. Good job!

lib/elliptic/curve/mont.js
sublimator: Or what about all just in base?

lib/elliptic/curve/base.js
sublimator: What do you mean?

sublimator: decodePoint should handle decoding just `x` (and 1 z) for mont curves. There's a test too. 

But maybe you mean something else?

lib/elliptic/curve/base.js
sublimator: That makes sense to me. I'm a big believer in blowing shit up :)

lib/elliptic/curve/mont.js
indutny: Meh, doesn't feel like OOP.

lib/elliptic/curve/base.js
indutny: I figured out that `decodePoint` is overriden in `mont`, but did it a bit later :( sorry.

lib/elliptic/curve/mont.js
sublimator: haha OK :)

I felt like it's easy to see all in one place, and it's going to be more code overall, with a bunch of repetition machinery. I'll have a crack at it.

sublimator: How would you do it? (You can push to my repo remember)

lib/elliptic/curve/base.js
sublimator: That does say something about the merits of keeping it all in one place :)

lib/elliptic/curve/mont.js
indutny: Well, I'd just take two branches of that `if` and put them into two methods. The preparation step might be put in some `_`-prefixed method in BasePoint.

sublimator:
``` I might add this
function intToBE(num, padTo) {
  var bytes = num.toArray();
  while (bytes.length < padTo)
    bytes.unshift(0);
  return bytes;
}
utils.intToBE = intToBE;
```

sublimator: A padTo on the bn might be nice :) ?

indutny: I think Pull Request function is working fine on bn.js repo ;)

lib/elliptic/curve/mont.js
sublimator: Hey, what's up :)

lib/elliptic/curve/mont.js
sublimator: hahaha, noted, will request pullage

lib/elliptic/curve/base.js
```diff
@@ -239,10 +240,51 @@ function BasePoint(curve, type) {
...
+
+  var len = this.p.byteLength();
+  if (bytes[0] === 0x04 && bytes.length - 1 === 2 * len) {
+    return this.point(bytes.slice(1, 1 + len),
+                      bytes.slice(1 + len, 1 + 2 * len));
+  } else if ((bytes[0] === 0x02 || bytes[0] === 0x03) &&
+              bytes.length - 1 === len) {
+    return this.pointFromX(bytes.slice(1, 1 + len), bytes[0] === 0x03);
+  }
+  throw new Error('Unknown point format');
```
sublimator: Guess should add a test for this

PR #49, opened 2015-07-16
[WIP] Add Point codec methods:

sublimator:
Add Point#eq
Add Point#encode
Add Point#encodeCompressed
Add Curve#decodePoint
Add overrides for Montgomery curves/points

Squash of #48

### DIFF
--- a/lib/elliptic/curve/base.js
+++ b/lib/elliptic/curve/base.js
@@ -2,10 +2,10 @@
 
 var bn = require('bn.js');
 var elliptic = require('../../elliptic');
-
-var getNAF = elliptic.utils.getNAF;
-var getJSF = elliptic.utils.getJSF;
-var assert = elliptic.utils.assert;
+var utils = elliptic.utils;
+var getNAF = utils.getNAF;
+var getJSF = utils.getJSF;
+var assert = utils.assert;
 
 function BaseCurve(type, conf) {
   this.type = type;
@@ -239,8 +239,44 @@ function BasePoint(curve, type) {
 }
 BaseCurve.BasePoint = BasePoint;
 
+BasePoint.prototype.eq = function eq(/*other*/) {
+  throw new Error('Not implemented');
+};
+
 BasePoint.prototype.validate = function validate() {
   return this.curve.validate(this);
+};
+
+BaseCurve.prototype.decodePoint = function decodePoint(bytes, enc) {
+  bytes = utils.toArray(bytes, enc);
+
+  var len = this.p.byteLength();
+  if (bytes[0] === 0x04 && bytes.length - 1 === 2 * len) {
+    return this.point(bytes.slice(1, 1 + len),
+                      bytes.slice(1 + len, 1 + 2 * len));
+  } else if ((bytes[0] === 0x02 || bytes[0] === 0x03) &&
+              bytes.length - 1 === len) {
+    return this.pointFromX(bytes.slice(1, 1 + len), bytes[0] === 0x03);
+  }
+  throw new Error('Unknown point format');
+};
+
+BasePoint.prototype.encodeCompressed = function encodeCompressed(enc) {
+  return this.encode(enc, true);
+};
+
+BasePoint.prototype._encode = function _encode(compact) {
+  var len = this.curve.p.byteLength();
+  var x = this.getX().toArray('be', len);
+
+  if (compact)
+    return [ this.getY().isEven() ? 0x02 : 0x03 ].concat(x);
+
+  return [ 0x04 ].concat(x, this.getY().toArray('be', len)) ;
+};
+
+BasePoint.prototype.encode = function encode(enc, compact) {
+  return utils.encode(this._encode(compact), enc);
 };
 
 BasePoint.prototype.precompute = function precompute(power) {
--- a/lib/elliptic/curve/mont.js
+++ b/lib/elliptic/curve/mont.js
@@ -5,6 +5,9 @@ var bn = require('bn.js');
 var inherits = require('inherits');
 var Base = curve.base;
 
+var elliptic = require('../../elliptic');
+var utils = elliptic.utils;
+
 function MontCurve(conf) {
   Base.call(this, 'mont', conf);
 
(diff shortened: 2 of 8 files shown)


lib/elliptic/curve/base.js
```diff
@@ -239,10 +240,51 @@ function BasePoint(curve, type) {
...
+
+  var len = this.p.byteLength();
+  if (bytes[0] === 0x04 && bytes.length - 1 === 2 * len) {
+    return this.point(bytes.slice(1, 1 + len),
+                      bytes.slice(1 + len, 1 + 2 * len));
+  } else if ((bytes[0] === 0x02 || bytes[0] === 0x03) &&
+              bytes.length - 1 === len) {
+    return this.pointFromX(bytes.slice(1, 1 + len), bytes[0] === 0x03);
+  }
+  throw new Error('Unknown point format');
```
sublimator: Should this method try and check for validity of the point ?

indutny: I wouldn't do it.

sublimator: I think I agree

lib/elliptic/curve/mont.js
```diff
@@ -147,6 +158,10 @@ Point.prototype.mulAdd = function mulAdd() {
   throw new Error('Not supported on Montgomery curve');
 };
 
+Point.prototype.eq = function eq(other) {
+  return this.getX().cmp(other.getX()) === 0;
```
sublimator: Does this need to check z is 1 or anything ?

indutny: To answer this question we need to asnwer: does `getX` normalize point?

sublimator: Indeed, I forgot about that :) And the normalized form puts z as 1 huh?

indutny: It is right below this function :)

sublimator: hahaha! That's funny

indutny: Its funny for you, but it scares me a lot. You patched this code all over, and didn't see a small function below. Now I will need to recheck everything :)

sublimator: Oh, come on, I was just thinking out loud, momentarily forgot the normalize
:)

sublimator: But yes, nothing wrong with rechecking :) :+1:

sublimator: If it calms you down, the `normalized form puts z as 1 huh?` was one of those questions that already had an answer.

indutny: Stay calm :) I was joking

sublimator: Keep calm and let @indutny handle it

sublimator:
Anything else you want changed here, or you want it squashed down?

sublimator:
@dcousens 
The party over here now :)

lib/elliptic/curve/base.js
```diff
@@ -239,10 +240,51 @@ function BasePoint(curve, type) {
...
+BasePoint.prototype._encode = function _encode(compact) {
+  var len = this.curve.p.byteLength();
+  var x = intToBE(this.getX(), len);
+
+  return compact ?
+      [ this.getY().isEven() ? 0x02 : 0x03 ].concat(x) :
+      [ 0x04 ].concat(x, intToBE(this.getY(), len)) ;
+};
+
+BasePoint.prototype.encode = function encode(compact, enc) {
```
sublimator: How do we feel about switching the order of the compact and enc args around ? In context of `encodeCompressed` ?

lib/elliptic/curve/base.js
```diff
@@ -239,10 +240,46 @@ function BasePoint(curve, type) {
...
+  var len = this.curve.p.byteLength();
+  var x = intToBE(this.getX(), len);
+
+  if (compact)
+    return [ this.getY().isEven() ? 0x02 : 0x03 ].concat(x);
+
+  return [ 0x04 ].concat(x, intToBE(this.getY(), len)) ;
+};
+
+BasePoint.prototype.encode = function encode(enc, compact) {
```
sublimator: I went ahead and swapped these around

lib/elliptic/utils.js
```diff
@@ -170,10 +170,12 @@ function intFromLE(bytes) {
...
-  while (bytes.length < padTo)
-    bytes.push(0);
-  return bytes;
+function intToLE(num, length) {
+  return num.toArray('le', length);
 }
 utils.intToLE = intToLE;
+
+function intToBE(num, length) {
+  return num.toArray('be', length);
```
indutny: Seems to be kind of small method, why not inline this code?

lib/elliptic/utils.js
```diff
@@ -170,10 +170,12 @@ function intFromLE(bytes) {
 }
 utils.intFromLE = intFromLE;
 
-function intToLE(num, padTo) {
-  var bytes = num.toArray('le');
-  while (bytes.length < padTo)
-    bytes.push(0);
-  return bytes;
+function intToLE(num, length) {
+  return num.toArray('le', length);
```
indutny: Same above. It is very slow, and not going to change much.

test/curve-test.js
```diff
@@ -162,3 +162,87 @@ describe('Curve', function() {
...
+      var encoded = p.encode('hex');
+      var decoded = curve.decodePoint(encoded, 'hex');
+      assert(decoded.eq(p));
+      assert.equal(encoded, definition.encoded);
+    }
+  }
+
+  var shortPointEvenY = {
+    coordinates: {
+      x: "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
```
indutny: Mixing `"` and `'`, let's use `'` everywhere.

lib/elliptic/utils.js
sublimator: Is there really any measurable performance cost?

test/curve-test.js
sublimator: Errrrk. Will use '

sublimator:
@indutny 

I inlined those methods, and fixed the quoting issues, along with some other lint issues. It might pay to enable linting etc for the tests eh ( when moving to semi-standard probably a good time) ?

indutny:
@sublimator there is a mental cost for them (at least for me). Linting tests makes sense to me!

indutny:
LGTM, please squash ;)

sublimator:
squashed

indutny:

sublimator:
cheers :)

PR #50, opened 2015-08-04
Improve utils.lazyComputed, and add instanbul for coverage

sublimator:


lib/elliptic/eddsa/signature.js
```diff
@@ -39,19 +39,19 @@ function Signature(eddsa, sig) {
   this._Sencoded = Array.isArray(sig.S) ? sig.S : sig.Sencoded;
 }
 
-lazyComputed(Signature, 'S', function S() {
```
sublimator: Note these suckers all using `function S()` :( :-1:

indutny:
Landed, thank you!

sublimator:
That was quick! Cheers :)

PR #55, opened 2015-09-29
Removed all cyclic dependencies.

originalgremlin:
Hi.  In my attempts to include this code in a React Native project I discovered a number of cyclic dependencies that prevented their default packager from completing.  With the liberal rewriting of a few requires all the cycles have been eliminated (according to pahen/madge).

These changes are all superficial.  All tests continue to pass.


dcousens:
Probably don't rebuild the `dist/` files yourself @originalgremlin,  it'll be done by the package maintainer.

dcousens:
Otherwise LGTM :+1:

indutny:
Hm... Doesn't LGTM. I use cyclic dependencies in all my project, as the way to deliver features to different parts of its internals. What about posting an issue at React Native instead? Fixing recursive requires is very simple.

dcousens:
@indutny on the plus side,  you get a `require` error if the module doesn't exist using this method compared to the `.utils` method.
The bundler might also be able to drop some modules if they aren't being used (as a pattern,  not necessarily in this case).

indutny:
@dcousens all modules are used :) And we have tests for obvious errors.

dcousens:
@indutny true,  but,  assuming there are no other reasons not too,  there is no *downside* to merging this PR?

indutny:
@dcousens the downside is that it sort of code style for me:

* https://github.com/indutny/node-spdy/blob/master/lib/spdy.js
* https://github.com/indutny/spdy-transport/blob/master/lib/spdy-transport.js
* https://github.com/indutny/asn1.js/blob/master/lib/asn1.js
* https://github.com/indutny/hash.js/blob/master/lib/hash.js
* https://github.com/indutny/json-pipeline/blob/master/lib/pipeline.js
* and the rest of my modules

It is a matter of principle for me :) Especially, considering that the root of the issue is not in this project anyway.

indutny:
@originalgremlin nevertheless, I appreciate your contribution! It just does seem to me that we are trying to fix the problem at the wrong place.

dcousens:
@indutny fair enough :ghost: 
@originalgremlin please reference the issue for `react-native` if you create one.

originalgremlin:
@indutny Yes, I agree that the react-native packager is far from complete and should deal with cyclic dependencies with more grace than a full lockup.

I don't really expect you to change all your codebases, considering that this is your preferred style.  To get my current code working the easiest and most direct way was to modify these few libraries rather than dig into the bowels of the packager.  I personally find cyclic dependencies to be a bit confusing to work with and prone to error.  They're certainly not in fashion in the wider community.  But for now I'll point my dependencies to my forks and start pestering the react-native folks.  Cheers.

dcousens:
> I personally find cyclic dependencies to be a bit confusing to work with and prone to error

Agreed, I personally don't find it easy to reason about their exact semantics, especially not at face value since it depends a lot on how Javascript works internally.

indutny:
Closing, sorry!

PR #58, opened 2015-10-24
better der encoding

calvinmetcalf:
p521 keys can have a total length greater then 127 bits leading to it being in long form, this covers that through we might just want to use asn1 instead of this.  Also adds tests for it but since p384 and p521 don't actually have malability issues rewrote the test to be more direct about what is being tested.

### DIFF
--- a/lib/elliptic/ec/signature.js
+++ b/lib/elliptic/ec/signature.js
@@ -23,34 +23,87 @@ function Signature(options, enc) {
 }
 module.exports = Signature;
 
+function Position() {
+  this.place = 0;
+}
+
+function getLength(buf, p) {
+  var initial = buf[p.place++];
+  if (!(initial & 0x80)) {
+    return initial;
+  }
+  var octetLen = initial & 0xf;
+  var val = 0;
+  for (var i = 0, off = p.place; i < octetLen; i++, off++) {
+    val <<= 8;
+    val |= buf[off];
+  }
+  p.place = off;
+  return val;
+}
+
+function rmPadding(buf) {
+  var i = 0;
+  var len = buf.length - 1;
+  while (!buf[i] && !(buf[i + 1] & 0x80) && i < len) {
+    i++;
+  }
+  if (i === 0) {
+    return buf;
+  }
+  return buf.slice(i);
+}
+
 Signature.prototype._importDER = function _importDER(data, enc) {
   data = utils.toArray(data, enc);
-  if (data.length < 6 || data[0] !== 0x30 || data[2] !== 0x02)
+  var p = new Position();
+  if (data[p.place++] !== 0x30) {
     return false;
-  var total = data[1];
-  if (1 + total > data.length)
+  }
+  var len = getLength(data, p);
+  if ((len + p.place) !== data.length) {
     return false;
-  var rlen = data[3];
-  // Short length notation
-  if (rlen >= 0x80)
+  }
+  if (data[p.place++] !== 0x02) {
     return false;
-  if (4 + rlen + 2 >= data.length)
+  }
+  var rlen = getLength(data, p);
+  var r = data.slice(p.place, rlen + p.place);
+  p.place += rlen;
+  if (data[p.place++] !== 0x02) {
     return false;
-  if (data[4 + rlen] !== 0x02)
+  }
+  var slen = getLength(data, p);
+  if (data.length !== slen + p.place) {
     return false;
-  var slen = data[5 + rlen];
-  // Short length notation
-  if (slen >= 0x80)
-    return false;
-  if (4 + rlen + 2 + slen > data.length)
-    return false;
+  }
+  var s = data.slice(p.place, slen + p.place);
+  if (r[0] === 0 && (r[1] & 0x80)) {
+    r = r.slice(1);
+  }
+  if (s[0] === 0 && (s[1] & 0x80)) {
+    s = s.slice(1);
+  }
 
-  this.r = new bn(data.slice(4, 4 + rlen));
-  this.s = new bn(data.slice(4 + rlen + 2, 4 + rlen + 2 + slen));
+  this.r = new bn(r);
+  this.s = new bn(s);
   this.recoveryParam = null;
 
   return true;
 };
+
+function constructLength(arr, len) {
+  if (len < 0x80) {
+    arr.push(len);
+    return;
+  }
+  var octets = 1 + (Math.log(len) / Math.LN2 >>> 3);
+  arr.push(octets | 0x80);
+  while (--octets) {
+    arr.push((len >>> (octets << 3)) & 0xff);
+  }
+  arr.push(len);
+}
 
 Signature.prototype.toDER = function toDER(enc) {
   var r = this.r.toArray();
@@ -63,8 +116,20 @@ Signature.prototype.toDER = function toDER(enc) {
   if (s[0] & 0x80)
     s = [ 0 ].concat(s);
 
-  var total = r.length + s.length + 4;
-  var res = [ 0x30, total, 0x02, r.length ];
-  res = res.concat(r, [ 0x02, s.length ], s);
+  r = rmPadding(r);
+  s = rmPadding(s);
+
+  while (!s[0] && !(s[1] & 0x80)) {
+    s = s.slice(1);
+  }
+  var arr = [ 0x02 ];
+  constructLength(arr, r.length);
+  arr = arr.concat(r);
+  arr.push(0x02);
+  constructLength(arr, s.length);
+  var backHalf = arr.concat(s);
+  var res = [ 0x30 ];
+  constructLength(res, backHalf.length);
+  res = res.concat(backHalf);
   return utils.encode(res, enc);
 };
--- a/test/ecdsa-test.js
+++ b/test/ecdsa-test.js
@@ -5,6 +5,7 @@ var hash = require('hash.js');
 describe('ECDSA', function() {
   function test(name) {
     it('should work with ' + name + ' curve', function() {
+      this.timeout(5000);
       var curve = elliptic.curves[name];
       assert(curve);
 
@@ -16,11 +17,16 @@ describe('ECDSA', function() {
         ]
       });
       var msg = 'deadbeef';
-
+      var keylen = 64;
+      if (name === 'p384') {
+        keylen = 96;
+      } else if (name === 'p521') {
+        keylen = 132
+      }
       // Get keys out of pair
       assert(keys.getPublic().x && keys.getPublic().y);
       assert(keys.getPrivate().length > 0);
-      assert.equal(keys.getPrivate('hex').length, 64);
+      assert.equal(keys.getPrivate('hex').length, keylen);
       assert(keys.getPublic('hex').length > 0);
       assert(keys.getPrivate('hex').length > 0);
       assert(keys.validate().result);
@@ -40,7 +46,7 @@ describe('ECDSA', function() {
 
       // key.sign(msg, options)
       var sign = keys.sign('hello', { canonical: true });
-      assert.notEqual(sign.toDER('hex'), keys.sign('hello').toDER('hex'));
+      assert(sign.s.cmp(keys.ec.nh) <= 0);
 
       // Load public key from compact hex
       var keys = ecdsa.keyFromPublic(keys.getPublic(true, 'hex'), 'hex');
@@ -66,6 +72,9 @@ describe('ECDSA', function() {
   }
   test('secp256k1');
   test('ed25519');
+  test('p256');
+  test('p384');
+  test('p521');
 
   describe('RFC6979 vector', function() {
     function test(opt) {


calvinmetcalf:
the failure is due to node 0.10 not having math.log2

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,74 @@ function Signature(options, enc) {
...
 
   return true;
 };
 
+function constructLength(arr, len) {
+  if (len < 0x80) {
+    arr.push(len);
+    return;
+  }
+  var octets = 1 + (Math.log2(len) >> 3);
```
indutny: Let's not use `Math.log2`, I'm trying to support as much browsers as possible.

indutny: Maybe we could use something like this https://github.com/indutny/bn.js/blob/master/lib/bn.js#L494-L520 ?

indutny: It could be even simpler than the one in bn.js as we just need number of octets, not the bit length.

calvinmetcalf: as it runs out Math.log2 is equivalent to Math.log(z)/Math.LN2 and those have pretty wide support

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,74 @@ function Signature(options, enc) {
...
 
   return true;
 };
 
+function constructLength(arr, len) {
+  if (len < 0x80) {
+    arr.push(len);
+    return;
+  }
+  var octets = 1 + (Math.log(len) / Math.LN2 >> 3);
```
indutny: `>>> 3` should be after parens

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,74 @@ function Signature(options, enc) {
...
+  if (data[p.place++] !== 0x02) {
     return false;
-  if (4 + rlen + 2 + slen > data.length)
+  }
+  var slen = getLength(data, p);
+  if (data.length !== slen + p.place) {
     return false;
+  }
+  var s = data.slice(p.place, slen + p.place);
+  if (!r[0] && (r[1] & 0x80)) {
```
indutny: `r[0] === 0`?

lib/elliptic/ec/signature.js
calvinmetcalf: the intent is  `1 + ((Math.log(len) / Math.LN2) >> 3);` or 0 - 255 turn into 1 and 256 to 2^16 turn into 2

lib/elliptic/ec/signature.js
calvinmetcalf: will fix

lib/elliptic/ec/signature.js
indutny: I thought that `/` has less priority than `>>>`, so it will do `>>>` first, and divide later. Turns out you are right, and this is correct statement.

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,74 @@ function Signature(options, enc) {
...
-    return false;
-  var total = data[1];
-  if (1 + total > data.length)
-    return false;
-  var rlen = data[3];
-  // Short length notation
-  if (rlen >= 0x80)
-    return false;
-  if (4 + rlen + 2 >= data.length)
+  var p = {};
```
indutny: Let's do a `function Position() { this.place = 0; }` for this, should help compiler to optimize this.

lib/elliptic/ec/signature.js
calvinmetcalf: I only know this because I looked it up when writing this

calvinmetcalf:
and green

test/ecdsa-test.js
```diff
@@ -5,6 +5,7 @@ var hash = require('hash.js');
 describe('ECDSA', function() {
   function test(name) {
     it('should work with ' + name + ' curve', function() {
+      this.timeout(5000);
```
indutny: Does it really take this long?

calvinmetcalf: it takes like 2800 on saucelabs so we could lower it to 4000

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
   return true;
 };
 
+function constructLength(arr, len) {
+  if (len < 0x80) {
+    arr.push(len);
+    return;
+  }
+  var octets = 1 + (Math.log(len) / Math.LN2 >> 3);
+  arr.push(octets ^ 0x80);
```
indutny: Should be `|`

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
+  while (true) {
```
indutny: I suggest to do a loop on `octets`, until it will be `0`.

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
+    if (octets === 1) {
+      arr.push(len & 0xff);
+      return;
+    }
+    octets--;
+    arr.push(len >> (octets << 3));
```
indutny: This is a bit incorrect, should be `(l en >>> (octets << 3)) & 0xff`

lib/elliptic/ec/signature.js
indutny: I would use `>>>` here too.

lib/elliptic/ec/signature.js
```diff
@@ -63,8 +104,19 @@ Signature.prototype.toDER = function toDER(enc) {
...
+    r = r.slice(1);
+  }
+  while (!s[0] && !(s[1] & 0x80)) {
+    s = s.slice(1);
+  }
+  var rarr = [ 0x02 ];
+  constructLength(rarr, r.length);
+  var sarr = [ 0x02 ];
+  constructLength(sarr, s.length);
+  var backHalf = rarr.concat(r, sarr, s);
```
indutny: can't we do it in a single array?

lib/elliptic/ec/signature.js
```diff
@@ -63,8 +104,19 @@ Signature.prototype.toDER = function toDER(enc) {
   if (s[0] & 0x80)
     s = [ 0 ].concat(s);
 
-  var total = r.length + s.length + 4;
-  var res = [ 0x30, total, 0x02, r.length ];
-  res = res.concat(r, [ 0x02, s.length ], s);
+  while (!r[0] && !(r[1] & 0x80)) {
+    r = r.slice(1);
```
indutny: This is quite unoptimal, let's find proper index and slice it after the loop.

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
-  if (slen >= 0x80)
-    return false;
-  if (4 + rlen + 2 + slen > data.length)
+  }
+  var slen = getLength(data, p);
+  if (data.length !== slen + p.place) {
     return false;
+  }
+  var s = data.slice(p.place, slen + p.place);
+  if (r[0] === 0 && (r[1] & 0x80)) {
```
indutny: Why not loop here?

indutny: Aaah, because it is well-formed here, ok...

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
+  var octetLen = initial & 0xf;
+  var byteLength = octetLen;
+  var val = buf[p.place + --byteLength];
+  var mul = 1;
+  while (byteLength > 0 && (mul *= 0x100)) {
+    val += buf[p.place + --byteLength] * mul;
+  }
+  p.place += octetLen;
+  return val;
+}
```
indutny: Let's add some newlines between the constructors, and also methods

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
+function Position() {
```
indutny: I think it kind of makes sense to move it above the `getLength`.

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
+function getLength(buf, p) {
+  var initial = buf[p.place++];
+  if (!(initial & 0x80)) {
+    return initial;
+  }
+  var octetLen = initial & 0xf;
+  var byteLength = octetLen;
+  var val = buf[p.place + --byteLength];
+  var mul = 1;
+  while (byteLength > 0 && (mul *= 0x100)) {
```
indutny: Perhaps you just need a `for` loop here?

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,76 @@ function Signature(options, enc) {
...
 module.exports = Signature;
 
+function getLength(buf, p) {
+  var initial = buf[p.place++];
+  if (!(initial & 0x80)) {
+    return initial;
+  }
+  var octetLen = initial & 0xf;
+  var byteLength = octetLen;
+  var val = buf[p.place + --byteLength];
```
indutny: Any reason it can't be a loop like this?

```javascript
var val = 0;
for (var i = 0, off = p.place; i < byteLength; i++, off++) {
  val <<= 8;
  val |= buf[off];
}
```

calvinmetcalf: done

lib/elliptic/ec/signature.js
```diff
@@ -21,37 +21,92 @@ function Signature(options, enc) {
...
+
+function getLength(buf, p) {
+  var initial = buf[p.place++];
+  if (!(initial & 0x80)) {
+    return initial;
+  }
+  var octetLen = initial & 0xf;
+  var byteLength = octetLen;
+  var val = 0;
+  for (var i = 0, off = p.place; i < byteLength; i++, off++) {
```
indutny: You don't need `byteLength` now, right?

lib/elliptic/ec/signature.js
```diff
@@ -21,37 +21,92 @@ function Signature(options, enc) {
...
+    val <<= 8;
+    val |= buf[off];
+  }
+  p.place += octetLen;
```
indutny: This could be just `p.place = off`

lib/elliptic/ec/signature.js
```diff
@@ -21,37 +21,92 @@ function Signature(options, enc) {
...
 };
 
+function constructLength(arr, len) {
+  if (len < 0x80) {
+    arr.push(len);
+    return;
+  }
+  var octets = 1 + (Math.log(len) / Math.LN2 >>> 3);
+  arr.push(octets | 0x80);
+  while (octets) {
```
indutny: Why not a `for` loop here?

lib/elliptic/ec/signature.js
```diff
@@ -21,37 +21,92 @@ function Signature(options, enc) {
...
+    if (octets-- === 1) {
+      arr.push(len & 0xff);
```
indutny: You should not really need this...

calvinmetcalf: is that taken care of latter?

indutny: Yup.

lib/elliptic/ec/signature.js
```diff
@@ -63,8 +118,21 @@ Signature.prototype.toDER = function toDER(enc) {
   if (s[0] & 0x80)
     s = [ 0 ].concat(s);
 
-  var total = r.length + s.length + 4;
-  var res = [ 0x30, total, 0x02, r.length ];
-  res = res.concat(r, [ 0x02, s.length ], s);
+  r = rmPadding(r);
+
```
indutny: No need in this.

lib/elliptic/ec/signature.js
```diff
@@ -21,37 +21,92 @@ function Signature(options, enc) {
   else
     this.recoveryParam = null;
 }
+
```
indutny: No need in this.

lib/elliptic/ec/signature.js
calvinmetcalf: seems like it would be more complicated since we have to do something with octets after we define it, we can make the loop a bit less complex

lib/elliptic/ec/signature.js
```diff
@@ -23,35 +23,85 @@ function Signature(options, enc) {
...
-  var total = data[1];
-  if (1 + total > data.length)
+  var p = new Position();
+  if (data[p.place++] !== 0x30) {
     return false;
-  var rlen = data[3];
-  // Short length notation
-  if (rlen >= 0x80)
+  }
+  getLength(data, p);
```
indutny: Let's validate this too.

calvinmetcalf: what do you mean?

indutny: This is a total length of the signature, its contents should have total length equal to this number.

calvinmetcalf: fixed

indutny:
Tests are failing :( After you will fix them, do you mind to squash the commits and give them an appropriate commit log? ;)

calvinmetcalf:
squashed and hopefully green

indutny:
Green! Thanks!

indutny:
Landed and released in 5.2.1

indutny:
Going to do a major bump with newest bn.js now.

PR #60, opened 2015-11-06
Use package.json files field instead of npmignore

TrySound:
With tests gziped package is 920kb. Without them just 38kb.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,0 @@
-benchmarks/
-node_modules/
-dist/*.js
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -1,6 +0,0 @@
-benchmarks/
-node_modules/
```
dcousens: `node_modules` should stay regardless

indutny:
Landed and released in 6.0.2, thank you!

PR #66, opened 2016-01-07
Added hex signature sample without private key

bolzon:
Just as requested ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,31 @@ var derSign = signature.toDER();
 console.log(key.verify(msg, derSign));
 ```
 
+When there's a public key and signature, but no private key:
+
+```javascript
+var EC = require('elliptic').ec;
+
+// Message to check
+var msg = 'hello world';
+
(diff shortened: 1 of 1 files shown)


indutny:
Perhaps we could use `keyFromPublic` there and leave aside the actual hex values? Also, let's omit that splitting example and say something more generic like:

```
signature MUST be either DER-encoded buffer or hex-string, or an object with two properties `r` and `s`, which should be hex strings.
```

bolzon:
Do you mean not to use real values?
Or just to use `keyFromPublic` and verify the message from key object?

I mean:

    var pub = '04bb1fab...';
    var ec = new EC('p256');
    var key = ec.keyFromPublic(pub, 'hex');
    console.log(key.verify(msg, signature);

indutny:
Yep, something like this. I think you could even merge it with the code chunk that we already have in readme, and just take `ec` that is already defined there.

I like how you put `pub` in your last comment.

Thank you!

Issue #68, opened 2016-01-14
custom nonce function


fanatid:
In https://github.com/cryptocoinjs/secp256k1-node/pull/57 we discussed using elliptic in v3.0.0, but it's required custom nonce function. Any ideas how it's can be implemented here? Wrapper for HmacDRBG? Does it make sense move HmacDRBG to separate module?

Relevant links:
[generating nonce in secp256k1-node v3.0.0](https://github.com/cryptocoinjs/secp256k1-node/blob/2ef419080f33b61fd845e8963c910d7cf0e3fd35/lib/ecdsa.js#L66)
[generating nonce in elliptic](https://github.com/indutny/elliptic/blob/58b2ee15346315d6f8c16313005f9421c2599df3/lib/elliptic/ec/index.js#L118)

indutny:
Submitted https://github.com/indutny/elliptic/pull/69 , please take a look ;)

PR #69, opened 2016-01-14
ecdsa: allow passing custom K for signing

indutny:
Fix: #68 

cc @fanatid

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -115,7 +115,9 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
   // Number of bytes to generate
   var ns1 = this.n.sub(new bn(1));
   do {
-    var k = new bn(drbg.generate(this.n.byteLength()));
+    var k = options.k ?
+        options.k() :
+        new bn(drbg.generate(this.n.byteLength()));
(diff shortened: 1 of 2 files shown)


fanatid:
Nice solution! What about pass attempt number to `options.k`?

btw, small improvement in `.sign`:
```js
  var bytes = this.n.byteLength();
  var bkey = key.getPrivate().toArray();
  for (var i = bkey.length; i < bytes; i++)
    bkey.unshift(0);
```
to
```js
var bkey = key.getPrivate().toArray(null, this.n.byteLength());
```
(same for nonce)

indutny:
Landed in 9fc874c . `options.k(iter)` should return `bn.js` instance indeed, it will be truncated if it is bigger than curve's `n` parameter. Otherwise it will try to use it as it is.

Please be aware that it is extremely important that the `k` should be never repeated when using the same private key and different messages. If there will ever be the same `k` - the private key could be easily extracted out of the signatures. This is the reason why I'm using HMAC-DRBG by default.

indutny:
Regarding optimizations - may I ask you to submit a PR with these changes?

dcousens:
@indutny awesome :+1:

fanatid:
done in #70

PR #70, opened 2016-01-16
lib: use toArray with options in EC.sign

fanatid:


indutny:
LGTM

indutny:

PR #71, opened 2016-01-20
Allow pass pers to drbg from ec.sign

fanatid:
In #68 I forgot mentioned that secp256k1-node v3.0.0 also need pass extra data to drbg.

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -105,7 +105,9 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
   var drbg = new elliptic.hmacDRBG({
     hash: this.hash,
     entropy: bkey,
-    nonce: nonce
+    nonce: nonce,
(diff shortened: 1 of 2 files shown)


indutny:
@fanatid just curious, what kind of data do you pass there? Does it follow any spec?

fanatid:
I want pass `Buffer` or `null`, but it always will go through `utils.toArray`.
I wasn't seen any spec for this (except RFC 6979 3.6 Additional Data), just trying follow bitcoin/secp256k1.

indutny:
Ok then... May I ask you to write a test for it? Maybe cross-check the output value with secp256k1 with it's pers data.

indutny:
Oh, also... please let me know when you will make changes. Github does not send notifications on new commits in PR.

fanatid:
@indutny updated!

indutny:

PR #72, opened 2016-01-25
lib: fix recoveryParam in ECDSA Signature

fanatid:
check only on `undefined`, because `Signature` is internal


indutny:
Landed, thank you!

fanatid:
@indutny can you publish on npm please?

indutny:
Done.

PR #73, opened 2016-01-28
ecdsa: use Shamir's trick in recoverPubKey

fanatid:
should be faster, but not sure, because g.mul uses fixedNafMul

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -199,7 +199,7 @@ EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
   // 1.6.1 Compute Q = r^-1 (sR -  eG)
   //               Q = r^-1 (sR + -eG)
   var rInv = signature.r.invm(n);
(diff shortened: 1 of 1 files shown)


indutny:
Indeed, I believe that current code should be faster. Please let me know if it is not.

dcousens:
Nice.

fanatid:
however, run benchmark (in secp256k1-node) with changed elliptic:
```
recover x 74.03 ops/sec ±1.70% (71 runs sampled)
recover(changed) x 82.06 ops/sec ±1.13% (67 runs sampled)
```
with mul and add we have 2 extra conversion from jacobian to affine coordinates
one conversion is 1redInvm + 4redMul

indutny:
Nice!

indutny:
Thank you

indutny:
LGTM

indutny:
While I'm going to land it anyway, it would be great to have the benchmark for this ;)

indutny:
Landed, thank you!

PR #74, opened 2016-01-28
travis: run with 0.10, 0.11, 0.12, 4, 5

fanatid:
@indutny don't you mind if I move Makefile to package.json:scripts?


indutny:
Sounds like a good idea!

indutny:
Landed in 5a67ace, thank you!

PR #75, opened 2016-01-29
remove Makefile

fanatid:
* update benchmark and tests code with style rules
* remove ancient dist
* remove Makefile and move lint to package.json:scripts

Do you need commands for building browser version of elliptic?


indutny:
Landed, thank you!

PR #77, opened 2016-02-07
readme: add EdDSA to API section

fanatid:


indutny:
@fanatid I wonder if we should say something about situations where `eddsa` should be used instead of `ec`?

fanatid:
Sorry, but I do not much familiar with `eddsa` as I would wanted (almost all my knowledge from [zoo picture](http://cr.yp.to/talks/2008.05.12/zoo.html))
What's you mean under situations? Performance and reduced risk of a random number generator attacks?

indutny:
No problem at all, I was distracted at the time of writing that comment, so please disregard it. LGTM, and landed. Thank you!

PR #89, opened 2016-05-19
Added build task for browsers

sorgloomer:
I've found a broken build task in gruntfile.js, and this is my humble attempt to fix it. Creates dist/elliptic.js and dist/elliptic.min.js using browserify and uglify. The dist folder is gitignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/dist/
 test/lib/
 node_modules/
 coverage/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,7 +8,17 @@ module.exports = function(grunt) {
         options: {
           transform: ['brfs']
         }
-      }
+      },
+      dist: {
+        files: {
+          'dist/elliptic.js': [ 'lib/elliptic.js']
+        },
+        options: {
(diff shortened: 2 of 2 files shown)


fanatid:
Why not add build task to package.json:scripts?
@indutny I think we can remove grunt if will use [zuul](https://github.com/defunctzombie/zuul), I did this yesterday with tape..

sorgloomer:
I can add a "browser": "grunt build" script to the package.json if you wish.

I think its a matter of taste. I dont like to use package.json scripts to tasks with more than a single command, because it's hard to maintain the multiplatform support for builds. The other day I saw scripts like "browserify && uglify" somewhere, that obviously didn't work on windows, had to run them manually. So i'd prefer grunt/gulp or at least simple js scripts (but not shell scripts or batch files) to "concatenate" tasks.

fanatid:
@sorgloomer you can use [npm-run-all](https://www.npmjs.com/package/npm-run-all)

indutny:
Yikes, sorry! Looks like I fixed this myself accidentally, without looking at this PR. Sorry!

The fix is here: 1b60ef4

PR #92, opened 2016-05-25
Fix standalone build

sorgloomer:
Specified the right entry module for browserify


dist/elliptic.js
```diff
@@ -3935,10 +3935,7 @@ utils.intFromLE = intFromLE;
   // BN
 
   function BN (number, base, endian) {
-    // May be `new BN(bn)` ?
```
indutny: There is no need to recompile it here, it will happen automatically on `npm version ...`. May I ask you to remove this file from this PR?

sorgloomer:
Sure, I will make a new one

sorgloomer:
Here is the new one https://github.com/indutny/elliptic/pull/93

PR #93, opened 2016-05-25
Fix standalone build

sorgloomer:


indutny:
LGTM

indutny:

PR #94, opened 2016-05-28
X25519 test vectors from RFC7748

philikon:
I'm trying to use elliptic.js for X25519 ECDH to interop with libsodium, but I'm having problems doing so. I can't even get elliptic.js to verify against the X25519 test vectors from RFC7748. What am I doing wrong?

### DIFF
--- a/test/ecdh-test.js
+++ b/test/ecdh-test.js
@@ -25,3 +25,27 @@ describe('ECDH', function() {
   test('ed25519');
   test('secp256k1');
 });
+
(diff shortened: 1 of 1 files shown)


indutny:
Sorry for delay, just got to it.

indutny:
I don't think that X25519 is supported natively by this repo, but it definitely can be implemented using elliptic curve primitives available here. Please let me know if you are (or know someone who may be) interested in contributing this. I can provide all necessary tips!

PR #96, opened 2016-05-30
Added primitive value comparison support for short curve points

kripod:


indutny:
Thank you for contribution! How does this work? May I ask you to add a test for it?

kripod:
This would add the ability to use operator-based point comparison. For example, in order to check whether any duplicate points are present in a data set, just use the following expression:

```js
const pointSet = []; // An array filled with points
new Set(pointSet)).size === pointSet.length
```

kripod:
Also, it would be possible to compare points without using `Point.prototype.eq()`.

Instead of `p1.eq(p2)`, we could use `p1 === p2`.

indutny:
Does it work, though? I'd love to see test where two points are created:

```js
const p1 = curve.pointFromX(/* same x as below */)
const p2 = curve.pointFromX(/* same x as above */);
assert(p1 === p2);
```

Anyway, needs tests! 😉

kripod:
Okay, I'll add test as soon as I'll have time to do so. If I recall correctly, this could also add support for finding a Point's index in an array using `.indexOf()`

indutny:
@kripod I just did a quick test, it doesn't look like adding `valueOf` helps there:

```js
function A(value) {
  this.value = value;
}

A.prototype.valueOf = function valueOf() {
  return { value: this.value };
};

var a = new A(1);
var b = new A(1);

console.log(a === b);  // false
console.log(a == b);   // false
console.log(a !== b);  // true
console.log(a != b);   // true
```

indutny:
I believe [SameValueZero](http://www.ecma-international.org/ecma-262/6.0/#sec-samevaluezero) is used to compare `Set` values to each other.

kripod:
What if you try returning a primitive (just `this.value`) in the `valueOf()` method? Comparisons should work like that according to MSDN.

kripod:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf

indutny:
@kripod this won't work with Set either as it is using SameValueZero comparison.

kripod:
Okay, I understand that now, thanks!
By the way, please check your Gitter! :)

indutny:
I have Gitter? :)

PR #100, opened 2016-08-23
Make the example complete, using curve25519

lovato:
Just to make it functional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,9 @@ console.log(key.verify(msg, signature));
 ### ECDH
 
 ```javascript
+var ECDH = require('elliptic').ec;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -119,6 +119,9 @@ console.log(key.verify(msg, signature));
 ### ECDH
 
 ```javascript
+var ECDH = require('elliptic').ec;
```
fanatid: Why `ECDH` instead not `EC` ?

indutny: Well, the top subtitle says `ECDH` ;)

indutny:
Thank you, landed!

PR #112, opened 2016-12-03
fix(key): Add error when only either x or y are provided

oberien:


indutny:
@oberien what about montgomery curves? They don't really have `y`...

oberien:
You are completely correct. For montgomery public keys the point only has an x coordinate while y is `undefined` which resulted in the check failing. I added code handling montgomery and weierstrass curves separately.

indutny:
Fixed minor nits and landed. Thank you!

PR #114, opened 2016-12-03
fix(curves): Set b to 1 for curve25519

oberien:
The montgomery parameters for curve25519 define b to be 1. Even though
this library doesn't use it for calculations, for consistency reasons
setting it to the standartized value improves readability. Also using it
in calculations outside of the library it resulted in problems and
errors, which will be fixed with this change.


dcousens:
@oberien maybe add a verification test?

oberien:
`b` is never used in any calculation for montgomery curves. If it was used, every calculation with it being 0 would have produced Infinity. The only useful test I could do would be to do calculations with b=0 and then with b=1 and expect the results to be equal. The problem with this is that if b was actually used at a later point, these tests would break.

indutny:
Thank you!

Issue #120, opened 2017-01-27
Use ECDH result as seed for new iteration


JamesCullum:
I want to write a tool that allows ECDH between more than two parties. For this I will need to use a key's derive function on a previous derivation, see [here](http://crypto.stackexchange.com/questions/3682/how-can-the-diffie-hellman-key-exchange-be-extended-to-three-parties). However, derivate returns a BN object instead of a key pair.

I tried to use the result of derivate to create a new key pair. If I import it as a public key using `keyFromPublic` it has coordinates close to zero, hence I assume it is wrong. If I use `keyFromPrivate` the console output looks like a legit key but the result is wrong.

Find my sample script below with its incorrect output.
```
var ECDH = require('elliptic').ec;
var ec = new ECDH('curve25519');

var A = ec.genKeyPair(), B = ec.genKeyPair(), C = ec.genKeyPair();

var AB = ec.keyFromPrivate(A.derive(B.getPublic()));
var BC = ec.keyFromPrivate(B.derive(C.getPublic()));
var CA = ec.keyFromPrivate(C.derive(A.getPublic()));

var ABC = C.derive(AB.getPublic());
var BCA = A.derive(BC.getPublic());
var CAB = B.derive(CA.getPublic());

console.log(ABC.toString(16));
console.log(BCA.toString(16));
console.log(CAB.toString(16));
```

Output

> 2976da3391bca687aaaa1ead5eccabdf2b280ffb71b51bfe3ded6ed76735e6ba
> 2b2e10b2e06550180b4939bffabd862cdd86f34c7e48e566aa75e002e435998c
> 54261b628319e3d1f4dfbf5402903a5f212e586a98d8671ecb951cdf747ce8dd

JamesCullum:
I'm not sure if I do something wrong or if there is no way of doing ECDH with multiple parties using your library.

fanatid:
`.derive` return X coordinate of multiplication, but you need a point!
fixed example:
```js
var ECDH = require('elliptic').ec;
var ec = new ECDH('curve25519');

var A = ec.genKeyPair(), B = ec.genKeyPair(), C = ec.genKeyPair();

var AB = A.getPublic().mul(B.getPrivate())
var BC = B.getPublic().mul(C.getPrivate())
var CA = C.getPublic().mul(A.getPrivate())

var ABC = AB.mul(C.getPrivate())
var BCA = BC.mul(A.getPrivate())
var CAB = CA.mul(B.getPrivate())

[...]

PR #121, opened 2017-01-27
doc: Add extra example for ECDH

fanatid:
Example from https://github.com/indutny/elliptic/issues/120

Maybe we should move examples to test folder?


indutny:
Thank you! It may be a good idea to move it to `examples`.

PR #130, opened 2017-06-05
Remove unreachable code

dlakata:
Found via a typescript error


indutny:
Thank you, and sorry for very long delay.

PR #133, opened 2017-07-26
No longer requires package.json for version number

lorem--ipsum:
Referring to https://github.com/indutny/elliptic/issues/106

Issue #106, opened 2016-09-20
Browserified package contains full package path


scholtzm:
```
$ npm list elliptic
└─┬ browserify@13.0.1
  └─┬ crypto-browserify@3.11.0
    └─┬ browserify-sign@4.0.0
      └── elliptic@6.3.1
```

After creating bundle via browserify, full package path is exposed directly in the bundle.

This is caused by 2 things:
1. [Requiring package.json.](https://github.com/indutny/elliptic/blob/master/lib/elliptic.js#L5)
2. Once the package is installed, extra fields are added into package.json such as `_where` - this value contains full path to parent/dependent package.

It's not a huge deal but disclosing folder structure is far from ideal.

fanatid:
@scholtzm thank you that pointed on this. Personally I ever never thought that bundle can contain personal information. Any ideas how this can be solved? (Except doubling version in two places)

fanatid:
@scholtzm I found solution for you.
Check https://github.com/nolanlawson/package-json-versionify or https://github.com/fanatid/browserify-package-json
(but you need pass option global equal true to transform)

We also can use https://github.com/webpro/versionify but I think it's shitty solution.

@indutny we can close this issue.

scholtzm:
Sorry for not replying. As a quick fix, I just added a post-process step which would remove the file path from the final bundle we are shipping.

Seems like `package-json-versionify` is spot on solution for this problem and I'll use it. Thanks.

fanatid:
Be careful with `package-json-versionify` because this package remove all except `version`.

lorem--ipsum:
This is still an issue, we're using ellptic through webpack and `elliptic/lib/elliptic.js:5` requires the package.json file to access its version number, which results in  packaging the entire file in the output JavaScript. I don't think it's a webpack/browserify issue though...

PR #140, opened 2017-12-07
No longer requires package.json

DevSide:
Because #133 was not merged.
Issue #106

I used **browserify-versionify** instead of **package-json-versionify** which does too much stuff and produces a very short module.


dcousens:
NACK,  you add the complexity of another dependency.

Compared to the import of __existing code__,  with the eventual burden of only 50 LOC in your browserify build?

DevSide:
What do you mean ?

dcousens:
@DevSide I don't use browserify dist... wouldn't importing this in `node` force the `lib` import to fail?

dcousens:
After reading into #106... this does seem a bit cludgy.
Additionally,  how would this work without using the `dist/` for import?

DevSide:
What's the use case of having the version exportable ? Could we just add append a comment with it in dist generated files ?

DevSide:
Could we just append a comment in dist generated files ?

Issue #142, opened 2017-12-25
Curve25519 ECDH doesn't produce test vectors


mahrud:
I'm working on implementing ed25519/curve25519 for OpenPGP.js. EdDSA using ed25519 works fine with some modifications to this library, however, I haven't been able to get ECDH using curve25519 to work. I found a few issues related to key generation that I'll include in a pull request soon, but there's a bigger problem that the implementation doesn't produce correct test vectors from [RFC7748](https://tools.ietf.org/html/rfc7748#section-5-2):

```
   Input scalar:
     4b66e9d4d1b4673c5ad22691957d6af5c11b6421e0ea01d42ca4169e7918ba0d
   Input scalar as a number (base 10):
     35156891815674817266734212754503633747
     128614016119564763269015315466259359304
   Input u-coordinate:
     e5210f12786811d3f4b7959d0538ae2c31dbe7106fc03c3efc4cd549c715a493
   Input u-coordinate as a number (base 10):
     88838573511839298940907593866106493194
     17338800022198945255395922347792736741
   Output u-coordinate:
     95cbde9476e8907d7aade45cb4b873f88b595a68799fa152e6f8f7647aac7957
```

Sample code:
```
var curve = require('elliptic').ec('curve25519'),
    BN = require('bn.js'),
    util = require('openpgp').util;

function rev(h) {
  return h.match(/../g).reverse().join('');
}

k = curve.keyFromPrivate(rev('4b66e9d4d1b4673c5ad22691957d6af5c11b6421e0ea01d42ca4169e7918ba0d'))
k.priv.eq(new BN('4dba18799e16a42cd401eae021641bc1f56a7d959126d25a3c67b4d1d4e96648', 16))
u = util.hex2Uint8Array('40'+rev('e5210f12786811d3f4b7959d0538ae2c31dbe7106fc03c3efc4cd549c715a493'))
p = curve.keyFromPublic(u)
p.pub.getX().eq(new BN('13a415c749d54cfc3e3cc06f10e7db312cae38059d95b7f4d3116878120f21e5', 16))
util.hexidump(k.derive(p.pub).toArray()) ===
  rev('95cbde9476e8907d7aade45cb4b873f88b595a68799fa152e6f8f7647aac7957')
```

mahrud:
I may have identified the source of the problem, though I'm not sure if this is the only cause:

[...]

PR #144, opened 2018-01-05
Fixes Curve25519/Ed25519 for use in OpenPGP.js

mahrud:
This pull request fixes point encoding problems for X25519 curves, fixes key generation for Curve25519, and adds a key generation function for Ed25519. These are necessary fixes needed for adding ECC support to OpenPGP.js in this pull request: https://github.com/openpgpjs/openpgpjs/pull/618 .

### DIFF
--- a/lib/elliptic/ec/key.js
+++ b/lib/elliptic/ec/key.js
@@ -38,6 +38,7 @@ KeyPair.fromPrivate = function fromPrivate(ec, priv, enc) {
   });
 };
 
+// TODO: should not validate for X25519
 KeyPair.prototype.validate = function validate() {
   var pub = this.getPublic();
 
@@ -51,13 +52,7 @@ KeyPair.prototype.validate = function validate() {
   return { result: true, reason: null };
 };
 
-KeyPair.prototype.getPublic = function getPublic(compact, enc) {
-  // compact is optional argument
-  if (typeof compact === 'string') {
-    enc = compact;
-    compact = null;
-  }
-
+KeyPair.prototype.getPublic = function getPublic(enc, compact) {
   if (!this.pub)
     this.pub = this.ec.g.mul(this.priv);
 
@@ -77,9 +72,17 @@ KeyPair.prototype.getPrivate = function getPrivate(enc) {
 KeyPair.prototype._importPrivate = function _importPrivate(key, enc) {
   this.priv = new BN(key, enc || 16);
 
-  // Ensure that the priv won't be bigger than n, otherwise we may fail
-  // in fixed multiplication method
-  this.priv = this.priv.umod(this.ec.curve.n);
+  // For Curve25519/Curve448 we have a specific procedure.
+  // FIXME for Curve448
+  if (this.ec.curve.type === 'mont') {
+    var one = this.ec.curve.one;
+    var mask = one.ushln(255 - 3).sub(one).ushln(3);
+    this.priv = this.priv.or(one.ushln(255 - 1));
+    this.priv = this.priv.and(mask);
+  } else
+    // Ensure that the priv won't be bigger than n, otherwise we may fail
+    // in fixed multiplication method
+    this.priv = this.priv.umod(this.ec.curve.n);
 };
 
 KeyPair.prototype._importPublic = function _importPublic(key, enc) {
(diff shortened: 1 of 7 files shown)


mahrud:
@indutny have you had a chance to review this?
I didn't add any tests here since it still doesn't produce test vectors from #142, but there are new tests in the `ecc_nist` branch of OpenPGP.js that incorporate changes this PR.

lib/elliptic/ec/key.js
```diff
@@ -51,13 +52,7 @@ KeyPair.prototype.validate = function validate() {
...
 };
 
-KeyPair.prototype.getPublic = function getPublic(compact, enc) {
-  // compact is optional argument
-  if (typeof compact === 'string') {
-    enc = compact;
-    compact = null;
-  }
-
+KeyPair.prototype.getPublic = function getPublic(enc, compact) {
```
indutny: This is going to be a major semver change.

lib/elliptic/eddsa/index.js
```diff
@@ -74,6 +79,23 @@ EDDSA.prototype.keyFromSecret = function keyFromSecret(secret) {
...
+  var drbg = new HmacDRBG({
+    hash: this.hash,
+    pers: options.pers,
+    persEnc: options.persEnc || 'utf8',
+    entropy: options.entropy || elliptic.rand(this.hash.hmacStrength),
+    entropyEnc: options.entropy && options.entropyEnc || 'utf8',
+    nonce: this.curve.n.toArray()
+  });
+
+  return this.keyFromSecret(drbg.generate(32));
```
indutny: Let's not hard-code `32` here.

lib/elliptic/eddsa/key.js
```diff
@@ -17,11 +17,18 @@ var cachedProperty = utils.cachedProperty;
...
-  this._secret = parseBytes(params.secret);
+  if (params.hasOwnProperty('secret'))
+    this._secret = parseBytes(params.secret);
   if (eddsa.isPoint(params.pub))
     this._pub = params.pub;
-  else
+  else {
     this._pubBytes = parseBytes(params.pub);
+    if (this._pubBytes && this._pubBytes.length === 33 &&
+        this._pubBytes[0] === 0x40)
```
indutny: Please add `{` `}`.

lib/elliptic/ec/key.js
mahrud: Do you know when you'd like to release v7.0.0? I can think of some other possibly breaking cleanups that might be worthwhile to do.

For instance, I think it would make sense to remove dependency on `hmac-drbg` and `brorand` and instead pass a callback function that returns randomness. This one is definitely breaking, but there are some other things that I can't quite remember off the top of my head.

lib/elliptic/eddsa/key.js
mahrud: Only here?

indutny: Yeah, current code style allows omitting braces if the `if` statement and its body fits into two lines total (the body must always be on the next line).

dkg:
what's the status on this pull request?  it'd be nice to get it included so that openpgp.js doesn't use a fork :)

mahrud:
@dkg sorry about the delay. I think it's ready.

mahrud:
Is there anything else I should do here?

lib/elliptic/ec/key.js
chjj: Don't know about this one. This breaks node.js convention: `enc` is typically the last argument. It's also going to be a huge pain replacing all the code which relies on the old parameter order.

Agree with removing the brorand dependency though.

mahrud: I think consistency is more important here. Also, I've done the work of replacing the code which relies on the old order in this project (which was minimal), and other projects can easily fix it whenever they update the vendored version of elliptic.

chesnokovilya:
@indutny Hey! So what prevents this PR from merging? Is it order of parameters in the line? `KeyPair.prototype.getPublic = function getPublic(enc, compact)`
I think in general brainpool and curve25519 support could be good for this library!

mahrud:
@chjj, I haven't used or followed the changes to this repository since last April, so I'm not confident I can add the conversion method you requested. I do remember a lot of pain dealing with conflicting endian-ness in different RFCs, and I kept that in mind when making this PR and added notes or references to relevant RFCs when necessary.

With regards to the order of `enc` and `compact`, my opinion is that checking the type of the first input and switching the order of inputs if it is a string, as it is currently, should be avoided.
https://github.com/indutny/elliptic/blob/ee7970b92f388e981d694be0436c4c8036b5d36c/lib/elliptic/ec/key.js#L53-L58
That said, if it is the only thing keeping this PR from being merged, then please feel free to amend the PR as you see fit.

I rebased the PR and brought it up to date with the version that openpgp.js uses. All tests are green.

mahrud:
Should I close this PR? Not sure why it's still open, despite being approved. If the changes have already been applied elsewhere, please feel free to close it.

mahrud:
I don't think so. It just was never merged...

Thijs-van-Drongelen:
Brainpool support would be nice

PR #146, opened 2018-01-25
edwards: generalise pointFromY

gnarula:
Generalise the pointFromY method to work for all Edwards Curves by
taking `c` and `a` parameters into account.


gnarula:
Added a test :)

indutny:
Thank you, landed!

gnarula:
ah, my bad. `assert.deepStrictEqual` was introduced in v.1.2.0. I could've just used `assert.strictEqual`. Sending in a new PR.

Issue #150, opened 2018-04-04
EC.recoverPubKey should take the input msg in hex (not in decimal)


nakov:
The `EC.recoverPubKey(msg, signature, j, enc)` method should take its `msg` parameter as **hex value** instead as decimal value: https://github.com/indutny/elliptic/blob/776c9b0e99832ab9fa6c5fa6f684c484ca5265b2/lib/elliptic/ec/index.js#L198

Thus, it will be consistent with `EC.sign(msg, key, enc, options)` and `EC.verify(msg, signature, key, enc)`.
- https://github.com/indutny/elliptic/blob/776c9b0e99832ab9fa6c5fa6f684c484ca5265b2/lib/elliptic/ec/index.js#L98
- https://github.com/indutny/elliptic/blob/776c9b0e99832ab9fa6c5fa6f684c484ca5265b2/lib/elliptic/ec/index.js#L155

PR #151, opened 2018-04-04
Bug fix: recoverPubKey -> take msg as hex

nakov:
The EC.recoverPubKey(msg, signature, j, enc) should take the input `msg` as **hex** number, not **decimal**.

See https://github.com/indutny/elliptic/issues/150


k1rill-fedoseev:
If I not mistaken, also we need to `_truncateToN` msg, so that line should be 
`var e = this._truncateToN(new BN(msg, 16));`

gianlucafrei:
It would be very cool if somebody could merge this fix.

aka-toxa:
hey could someone merge this?

PR #188, opened 2019-08-13
fix Point.prototype.toJSON for short curves

mvayngrib:
thanks @fanatid for helping me look into this


fanatid:
Short explanation:

p192, p224, curve25519, ed25519, secp256k1 use `Base reduction engine`
p256, p384, p521 use `Montgomery method engine`

BN-Red for curve created at https://github.com/indutny/elliptic/blob/v6.5.0/lib/elliptic/curve/base.js#L14

On serialization `BN#toJSON` called without checking red existence: https://github.com/indutny/bn.js/blob/master/lib/bn.js#L537-L539

While `Base reduction engine` call just remove `red` property: https://github.com/indutny/bn.js/blob/master/lib/bn.js#L3442-L3446

`Montgomery method engine` makes manipulations: https://github.com/indutny/bn.js/blob/master/lib/bn.js#L3478-L3482

As result if point with Montgomery method will be serialized result will be not same on deserialization.

jprichardson:
@indutny @fanatid can we get this merged? Anything more we can help with to move this forward?

fanatid:
LGTM

PR #203, opened 2019-10-05
Apply nonce bit-length mitigation to stop timing leakage.

J08nY:
This should be a mitigation for the Minerva attack. See https://minerva.crocs.fi.muni.cz for more info. Some tests would be nice to see this is really constant time (and wasn't before), but I am not that well versed in JavaScript.

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -125,7 +125,17 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (k.cmpn(1) <= 0 || k.cmp(ns1) >= 0)
       continue;
 
-    var kp = this.g.mul(k);
+    // Fix the bit-length of the random nonce,
+    // so that it doesn't leak via timing.
+    // This does not change that ks = k mod n
+    var ks = k.add(this.n);
+    var kt = ks.add(this.n);
+    var kp;
+    if (ks.bitLength() == this.n.bitLength())
+      kp = this.g.mul(kt);
+    else
+      kp = this.g.mul(ks);
+
     if (kp.isInfinity())
       continue;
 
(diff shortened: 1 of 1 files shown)


J08nY:
Hi @indutny, is there any movement on this? This project is used by many others and remains likely vulnerable and not fixed months after our disclosure.

Rob-pw:
Thanks for the PR, just wandering by

lib/elliptic/ec/index.js
```diff
@@ -125,7 +125,17 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
...
       continue;
 
-    var kp = this.g.mul(k);
+    // Fix the bit-length of the random nonce,
+    // so that it doesn't leak via timing.
+    // This does not change that ks = k mod n
+    var ks = k.add(this.n);
+    var kt = ks.add(this.n);
+    var kp;
+    if (ks.bitLength() == this.n.bitLength())
```
indutny:
```suggestion
    if (ks.bitLength() === this.n.bitLength()) {
```

lib/elliptic/ec/index.js
```diff
@@ -125,7 +125,17 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
...
+      kp = this.g.mul(kt);
+    else
```
indutny:
```suggestion
    } else {
```

lib/elliptic/ec/index.js
```diff
@@ -125,7 +125,17 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
...
+      kp = this.g.mul(ks);
+
```
indutny:
```suggestion
      }

```

J08nY:
> LGTM. Thank you for submitting this and sorry for delay. There are few minor changes that I'd like to request before landing it, though!

Fixed those. Please test the changes (and timing).

indutny:
Hey, I've went ahead and applied alternative "fix": ec735ed . However, as bad as it sounds it must be said that this library was never meant to be timing-leak free. Architecture of elliptic and bn.js was designed to provide the fastest possible pure JS implementation of elliptic curve so that it would be usable in browsers. Patching this one particular loophole doesn't mean that there are no other side-channels (in fact, there are many through the source code).

PR #206, opened 2019-11-20
fix: Apply nonce bit-length mitigation to stop timing leakage.

evertonfraga:
This is a follow-up PR to https://github.com/indutny/elliptic/pull/203, created by @J08nY that adds the following:
- [x] code style error, as seen on [travis build](https://travis-ci.org/indutny/elliptic/jobs/600683572)
- [x] add built change to `/dist`
- [x] bumps version to `6.5.2`

I could've first PR'd to @J08nY's fork and then have it here, but wanted to take the shortcut, for simplicity sake.

Travis job for sauce labs testing will probably fail due to the lack of an API key for builds in external forks. It's listed as an optional test, so might not be that important.

### DIFF
--- a/dist/elliptic.js
+++ b/dist/elliptic.js
@@ -2288,7 +2288,18 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
     if (k.cmpn(1) <= 0 || k.cmp(ns1) >= 0)
       continue;
 
-    var kp = this.g.mul(k);
+    // Fix the bit-length of the random nonce,
+    // so that it doesn't leak via timing.
+    // This does not change that ks = k mod n
+    var ks = k.add(this.n);
+    var kt = ks.add(this.n);
+    var kp;
+    if (ks.bitLength() === this.n.bitLength()) {
+      kp = this.g.mul(kt);
+    } else {
+      kp = this.g.mul(ks);
+    }
+
     if (kp.isInfinity())
       continue;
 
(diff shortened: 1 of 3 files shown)


indutny:
Hey, I've went ahead and applied alternative "fix": ec735ed . However, as bad as it sounds it must be said that this library was never meant to be timing-leak free. Architecture of elliptic and bn.js was designed to provide the fastest possible pure JS implementation of elliptic curve so that it would be usable in browsers. Patching this one particular loophole doesn't mean that there are no other side-channels (in fact, there are many through the source code).

evertonfraga:
Fedor, I appreciate the quick response and action!

I am no specialist but I hope the mentioned changes satisfies the concerns raised by @J08nY in full.

Cheers :)

J08nY:
I am not sure the mentioned commit fixes the issue in scalar mult. As you are using incomplete formulas: https://github.com/indutny/elliptic/blob/master/lib/elliptic/curve/short.js#L367 which short-circuit on input of the point at infinity, so that just changing the length of the NAF array by prepending zeros will not make scalar mult not leak the bit-length of the NAF representation. Since all the leading zeros will just be skipped when adding/doubling the point at infinity, and the first non-zero NAF digit will start to take time.

indutny:
This is a good point. I guess I'll have to revert this and apply your original patch. Sorry!

nivida:
@indutny Btw. & off-topic: Is it also possible to introduce release notes and for example a [CHANGELOG](https://keepachangelog.com/en/1.0.0/)?

PR #232, opened 2020-08-14
getKeyRecoveryParam method reusing variable `e`, type incorrect

bradlhart:
`@types/elliptic` states that `getKeyRecoveryParam` `e` variable is `Error | undefined`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a9f3af6e4416678bad8027c15627683621f532f1/types/elliptic/index.d.ts#L219

However, other than within the scope of the `catch (e) {...}` where it is `Error | undefined`, the type is more likely either `BNInput` or one of the types `BNInput` represents as it is passed into `recoverPubKey`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a9f3af6e4416678bad8027c15627683621f532f1/types/elliptic/index.d.ts#L213

Likely why the `@types/elliptic` has the type wrong is due to `e` being redeclared in the `catch` statement.  This PR is just to fix the redeclaration.

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -223,7 +223,7 @@ EC.prototype.recoverPubKey = function(msg, signature, j, enc) {
   return this.g.mulAdd(s1, r, s2);
 };
 
-EC.prototype.getKeyRecoveryParam = function(e, signature, Q, enc) {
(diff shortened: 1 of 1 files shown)


irfan798:
I'm having the same issue with typescript, thank you for pointing that out @bradlhart

PR #244, opened 2021-01-26
Fixes a twist attack made possible in secp256k1

kdenhartog:
This addresses a twist attack outlined here: https://github.com/christianlundkvist/blog/blob/master/2020_05_26_secp256k1_twist_attacks/secp256k1_twist_attacks.md

I confirmed that any point is possible to pass into the derive function as a public key in the first commit then implemented a fix for it in the second commit. It can be squashed together on merge/I can squash it to keep the commit log simple if that's preferred.


RonSherfey (commented):
review of changes

Issue #250, opened 2021-03-03
EDDSA.decodePoint should fails on ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff


mahnunchik:
Working with `ed25519` curve I've faced with strange behaviour: `decodePoint` doesn't fails on points out of field.

```js
import elliptic from 'elliptic';

const ec = new elliptic.eddsa('ed25519');

const point = ec.decodePoint('ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff');
// doesn't throw Error('invalid point')

point.validate();
// returns true
```

What do you think?
* Is it a bug?
* Is it an implementation feature?
* Or I don't understand the logic behind the `decodePoint`?

mahnunchik:
I'm working on implementing the following method on js: https://github.com/monero-project/monero/blob/v0.17.1.9/src/crypto/crypto-ops.c#L1334-L1424

There are several test cases: https://github.com/monero-project/monero/blob/v0.17.1.9/tests/crypto/tests.txt (rows starts with `check_key`).

The following test reports **false positive** results using `decodePoint` and `validate` methods:

```txt
✕ keyCheck 'edffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f' is valid 'false'
✕ keyCheck 'edffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' is valid 'false'
✕ keyCheck 'eeffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f' is valid 'false'
✕ keyCheck 'f0ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f' is valid 'false'
✕ keyCheck 'f0ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' is valid 'false'
✕ keyCheck 'f1ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f' is valid 'false'
✕ keyCheck 'f1ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' is valid 'false'
✕ keyCheck 'f2ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f' is valid 'false'
✕ keyCheck 'f2ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' is valid 'false'
[...]

PR #253, opened 2021-03-17
fix: decode ec point according to the spec

mahnunchik:
This PR includes `decodePoint` method according to the specification: https://tools.ietf.org/html/rfc8032#section-5.1.3

> If the resulting value is >= p, decoding fails.

There are some tests, if needed more tests can be added.

Related issue: https://github.com/indutny/elliptic/issues/250


mahnunchik:
@indutny @fanatid ping

PR #326, opened 2024-10-25
fix: signature verification due to leading zeros

indutny:
According to FIPS 186-5, section 6.4.2 ECDSA Signature Verification Algorithm, the hash of the message must be adjusted based on the order n of the base point of the elliptic curve:

    If log2(n) ≥ hashlen, set E = H. Otherwise, set E equal to
    the leftmost log2(n) bits of H.

Unfortunately because elliptic converts messages to BN instances the reported `byteLength()` for the message can be incorrect if the message has 8 or more leading zero bits.

Here we fix it by:

1. Counting leading zeroes in hex strings provided as messages
2. Counting all array entries in Array-like (e.g. Buffer) messages
3. Providing an `msgBitLength` option to both `.sign`/`.verify` to let user override the behavior

Original PR: https://github.com/indutny/elliptic/pull/322
Credit: @Markus-MS

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -78,8 +78,27 @@ EC.prototype.genKeyPair = function genKeyPair(options) {
   }
 };
 
-EC.prototype._truncateToN = function _truncateToN(msg, truncOnly) {
-  var delta = msg.byteLength() * 8 - this.n.bitLength();
+EC.prototype._truncateToN = function _truncateToN(msg, truncOnly, bitLength) {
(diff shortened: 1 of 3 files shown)


paulmillr (approved):

paulmillr:
Who would use `options.msgBitLength`, and what for?

indutny:
@paulmillr thanks for the review! I just wanted to provide an escape hatch for people whom this patch broke :sob:

indutny:
Published in 6.6.0. Thank you!

PR #337, opened 2025-09-05
ed25519: canonical enc/dec, hash binding, and subgroup checks

MegaManSec:
**Fix 1 — `decodePoint`: enforce 32 bytes; reject `y ≥ p`**
Require exactly 32 bytes, clear the MSB to recover `y`, and fail if the resulting `y` is ≥ field prime `p`. Ensures strict, non-malleable decoding. **Ref:** RFC 8032 §5.1.3.

**Fix 2 — `verify`: bind hash to actual points**
Re-encode `R` and `A` from the decoded points for the challenge hash instead of trusting attacker-controlled `Rencoded()/pubBytes()`. Prevents verification confusion/object spoofing.

**Fix 3 — `verify`: enforce subgroup membership**
Reject identity or small-order `A`/`R` by checking `P * n == ∞`. Closes cofactor-related attacks that can yield bogus verifications in multi-sig/aggregation flows.

Also add some tests.


MegaManSec:
some pocs: https://gist.github.com/MegaManSec/62366c77efb75ac2459a28c22a2d8eaa

PR #338, opened 2025-09-05
ec: Add stronger key validation in verify and fix private key hex encoding

MegaManSec:
* add public key validation in `EC.verify` (same curve, valid point, passes `validate()`)
* fix `KeyPair.getPrivate('hex')` to output zero-padded hex of correct length
* correct typo in `recoverPubKey` error message ("sencond" → "second")


rurban:
ping

Security relevant

PR #345, opened 2026-01-16
fix: preserve leading zeros in k-value during ECDSA signing (RFC 6979)

drzippie:
## Summary

- Fix critical security vulnerability where k values with leading zeros are incorrectly truncated during ECDSA signature generation, violating RFC 6979 step 3.2
- Pass raw DRBG output array directly to `_truncateToN` instead of converting to BN first, preserving correct byte length for truncation calculations

## Security Advisory

This PR addresses **[CVE-2025-14505](https://github.com/advisories/GHSA-848j-6mx2-7j84)** (GHSA-848j-6mx2-7j84)

- **Severity:** Low (CVSS v4: 2.9/10)
- **CWE:** CWE-1240 - Use of a Cryptographic Primitive with a Risky Implementation
- **Affected versions:** All versions ≤ 6.6.1

### Impact

The vulnerability could allow attackers to potentially derive the secret key if they could obtain both a faulty signature generated by a vulnerable version of Elliptic and a correct signature for the same inputs.

## The Problem

When generating k values for ECDSA signatures:
1. `drbg.generate(32)` returns a 32-byte array (e.g., `[0x00, 0x00, 0x12, ...]`)
2. `new BN(array)` strips leading zeros (BN stores only significant bytes)
3. `_truncateToN(k, true)` uses `msg.byteLength()` which returns 30 instead of 32
4. Truncation calculation is wrong: `delta = 240 - 256 = -16` instead of `0`

This could lead to incorrect signatures and potential private key exposure via cryptanalysis.

## The Fix

The `_truncateToN` function already handles arrays correctly at lines 86-89:
```javascript
} else if (typeof msg === 'object') {
  byteLength = msg.length;  // Uses actual array length, not significant bytes
  msg = new BN(msg, 16);
}
```

By passing the raw array from `drbg.generate()` directly to `_truncateToN`, we preserve the full byte count for correct truncation.

## Test plan

- [x] All existing tests pass (228 tests)
- [x] RFC 6979 test vectors still pass
- [x] Added regression tests for leading zero k value handling
- [x] Added test for deterministic signature generation

### DIFF
--- a/lib/elliptic/ec/index.js
+++ b/lib/elliptic/ec/index.js
@@ -151,10 +151,13 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
   var ns1 = this.n.sub(new BN(1));
 
   for (var iter = 0; ; iter++) {
-    var k = options.k ?
-      options.k(iter) :
-      new BN(drbg.generate(this.n.byteLength()));
-    k = this._truncateToN(k, true);
+    var k;
+    if (options.k) {
+      k = this._truncateToN(options.k(iter), true);
+    } else {
+      // Pass array directly to _truncateToN to preserve leading zero byte count
+      k = this._truncateToN(drbg.generate(this.n.byteLength()), true);
+    }
     if (k.cmpn(1) <= 0 || k.cmp(ns1) >= 0)
       continue;
 
--- a/test/ecdsa-test.js
+++ b/test/ecdsa-test.js
@@ -544,4 +544,41 @@ describe('ECDSA', function() {
       assert.equal(sig.recoveryParam, 1);
     });
   });
+
+  describe('Leading zero k value handling', function() {
+    it('should correctly sign when k has leading zeros', function() {
+      var ec = new elliptic.ec('secp256k1');
+      var key = ec.genKeyPair({ entropy: entropy });
+
+      // Create k with leading zeros - this tests the fix
+      var leadingZeroK = new Array(32);
+      leadingZeroK[0] = 0x00;
+      leadingZeroK[1] = 0x00;
+      for (var i = 2; i < 32; i++) {
+        leadingZeroK[i] = 0x42 + i;
+      }
+
+      var sign = key.sign('test message', {
+        k: function(iter) {
+          if (iter === 0) return new BN(leadingZeroK);
+          return new BN(1358);
+        },
+      });
+
+      assert(ec.verify('test message', sign, key));
+    });
+
+    it('should produce deterministic signatures with potential leading zero k', function() {
+      var ec = new elliptic.ec('secp256k1');
+      var key = ec.keyFromPrivate(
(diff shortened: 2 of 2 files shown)


minahussain (approved):

lib/elliptic/ec/index.js
```diff
@@ -151,10 +151,13 @@ EC.prototype.sign = function sign(msg, key, enc, options) {
...
-      options.k(iter) :
-      new BN(drbg.generate(this.n.byteLength()));
-    k = this._truncateToN(k, true);
+    var k;
+    if (options.k) {
+      k = this._truncateToN(options.k(iter), true);
+    } else {
+      // Pass array directly to _truncateToN to preserve leading zero byte count
+      k = this._truncateToN(drbg.generate(this.n.byteLength()), true);
+    }
```
dzikowski, line 160: I may miss something, but the only change I see here is the removal of `new BN(...)`. If that's actually the fix, I think the change might be minimal and the code less verbose:

```js
    var k = options.k ?
      options.k(iter) :
      drbg.generate(this.n.byteLength());
    k = this._truncateToN(k, true);
```

joselarroque-opp-fi (approved):

vpetrusevici:
Hi @indutny,
Can you review this one please? It's necessary to eliminate the vulnerability.

CezaryDanielNowak:
@drzippie, removing `BN(drbg.generate(...))` will introduce other security issues. I restored BN and optimized the if-block.

Please check https://github.com/drzippie/elliptic/pull/1/files

The library is still not secure unfortunately

ama-leanix:
Any update to mitigate this vulnerability, please? 🙏

Mehben:
Hi @indutny, 

This fix is blocking security compliance for several downstream projects. The vulnerability (CVE-2025-25473) is flagged by Dependabot/GitHub Advisory and there's no workaround since 6.6.1 is the latest release.

Would you be able to review and merge this? Happy to help test if needed.

ipfizz:
+1 on this fix. Verified the diff against current master - the change correctly routes DRBG output through the array-aware branch of `_truncateToN` (which uses `msg.length`, the same pattern already adopted for `msg` in 34c8534). Affected here too via the `georaster > worker-loader > webpack@4 > node-libs-browser > crypto-browserify > elliptic` chain. Would adopt this fix as soon as it merges. Adding a voice to the existing approvals.

DanielChutkowskiJMMJ (approved):

juan-dalmasso-clio (approved):

juan-dalmasso-clio:
Hi team! Is there any chance to get this one merged?

vavsab:
If you are tired of waiting you can override with our fork

npm
```json
"overrides": {
     "elliptic": "npm:@mapped/elliptic@^6.6.5"
}
```

yarn
```json
"resolutions": {
     "elliptic": "npm:@mapped/elliptic@^6.6.5"
}
```


Unfortunately, you need to repeat it in every leaf project - overrides do not propagate.

PR #351, opened 2026-06-26
fix: correctly handle computation of k with leading zeros (RFC 6979) — CVE-2025-14505

zuzukobaladze:
## Summary
ECDSA's per-iteration nonce `k` is generated as a fixed-width buffer of `this.n.byteLength()` bytes, but is then truncated with `this._truncateToN(k, true)`, which derives the bit length from `BN#byteLength()`. For values whose fixed-width representation has leading zero byte(s), those bytes are dropped, giving the wrong RFC 6979 §2.3.2 `bits2int` shift and an incorrect `k`.

This produces faulty signatures for the affected `k` values; combined with a correct signature over the same input, it can leak the private key (CVE-2025-14505 / GHSA-848j-6mx2-7j84).

The fix passes the full generated bit length explicitly, mirroring the existing message path (`options.msgBitLength`):

```js
k = this._truncateToN(k, true, this.n.byteLength() * 8);
```

`_truncateToN` itself is unchanged.

## Test
Adds a p521 regression test that pins `r` to the value derived from the correctly full-width-truncated `k`. A plain sign/verify round-trip cannot detect this bug, since a single signature is always self-consistent for whatever `k` was actually used. The test fails on `master` and passes with the fix.

## Note
This is an already-public advisory, so disclosure here is appropriate. Consumers who produced signatures with a vulnerable build should rotate affected keys.

Made with [Cursor](https://cursor.com)


Shivam-bsh:
Hi @zuzukobaladze 
When do you plan to merge this PR ?

